PR #44, opened 2026-06-12
Add ICU formatter support diagnostics

swernerx:
## Summary

- add a generic `IcuFormatterSupport` decision type
- add `validate_icu_formatter_support(...)` to turn consumer runtime support policies into standardized ICU diagnostics
- distinguish unsupported formatter kind from unsupported formatter style
- re-export the API from both `ferrocat-icu` and the umbrella `ferrocat` crate

Refs sebastian-software/palamedes#147

## Test Plan

- `cargo test -p ferrocat-icu`
- `cargo test --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings`

### DIFF
--- a/crates/ferrocat-icu/README.md
+++ b/crates/ferrocat-icu/README.md
@@ -16,6 +16,7 @@ Use this crate when you want the ICU-specific surface directly:
 - `validate_icu` for lightweight validation
 - `analyze_icu` for structured argument, formatter, plural, select, and tag summaries
 - `compare_icu_messages` for source/translation compatibility diagnostics
+- `validate_icu_formatter_support` for mapping consumer runtime support policies to diagnostics
 - `normalize_message_metadata` for progressive source-side metadata around `msgid + msgctxt`
 - `extract_argument_names` and `extract_tag_names` when tags should not be mixed with data arguments
 - `extract_variables`, `has_plural`, `has_select`, and related helpers for AST inspection
--- a/crates/ferrocat-icu/src/analysis.rs
+++ b/crates/ferrocat-icu/src/analysis.rs
@@ -75,6 +75,23 @@ pub struct IcuFormatter {
     pub style_kind: IcuStyleKind,
 }
 
+/// Consumer-defined support decision for an ICU formatter.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum IcuFormatterSupport {
+    /// The formatter kind and style are supported by the consumer.
+    Supported,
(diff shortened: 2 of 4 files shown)


swernerx:
Reviewed in the context of sebastian-software/palamedes#147.

**Overall: good cut, mergeable.** The boundary is right — ferrocat doesn't guess what a runtime supports, the consumer passes a policy callback and ferrocat standardizes the diagnostics. The new codes (`icu.unsupported_formatter_kind/_style`) fit the existing naming scheme, all three decision paths are tested, CI is green across Linux/macOS/Windows + MSRV + lint. Landing the diagnostics side before the runtime side is the right order: until the Palamedes runtime learns `{x, date}` / number styles, it can declare every formatter kind unsupported and turn runtime surprises into compile-time diagnostics.

**Before merge — retitle for release-please.** This repo releases via release-please and the squash commit will be the PR title. `Add ICU formatter support diagnostics` has no conventional-commit prefix, so it triggers no minor bump and lands in no changelog. Suggest `feat(icu): add formatter support diagnostics`.

**Worth considering:**

1. `validate_icu_formatter_support` calls `analyze_icu` internally. Callers that already hold an analysis (the Palamedes compile pipeline analyzes each message for metadata anyway) pay a second AST walk per message. Suggestion: define the core function on `&IcuAnalysis` and keep the current `&IcuMessage` signature as a thin convenience. Matters once catalogs reach the 10k+ range (sebastian-software/palamedes#154).
2. The diagnostic texts use `{:?}` for the kind ("unsupported formatter kind `List`"). Palamedes plugins surface `diagnostic.message` verbatim to developers, where the ICU keyword (`list`, `number`) would read better than the Rust enum variant. A `Display` impl on `IcuArgumentKind` would cover this everywhere.

**Minor:**

- `UnsupportedStyle` with `style: None` renders "style `<none>`" — robust, but the docs should say when a consumer is expected to choose `UnsupportedStyle` vs `UnsupportedKind` (a missing style reads more like a default-style/kind decision).
- Two nice-to-have tests: a mixed report (several formatters, different decisions in one message), and one pinning down that plural/select arguments do **not** flow through the formatter callback — that boundary is worth a regression guard.

**Follow-up on the Palamedes side after release:** bump the pin to 0.13 (cargo 0.x semantics — `^0.12` won't match), wire the policy into catalog compile/audit, and keep the Rust-side policy and the actual TS runtime (`packages/core/src/messageFormat.ts`) from drifting apart — the support matrix needs a single source of truth, otherwise the diagnostics lie in one direction or the other as runtime support grows.

PR #79, opened 2026-06-12
feat(icu): validate formatter support from analysis

### LINKED ISSUE #46: icu: follow-ups from formatter-support review (#44)
PR #44 merged before review wrapped up — collecting the open points from https://github.com/sebastian-software/ferrocat/pull/44#issuecomment-4694770072 so they don't get lost. The release-please concern is resolved (merge commit got the `feat:` prefix, 0.13.0 is in release PR #45).

**1. Avoid the double AST walk.**
`validate_icu_formatter_support` calls `analyze_icu` internally. Callers that already hold an `IcuAnalysis` (the Palamedes compile pipeline analyzes each message for metadata anyway) pay a second traversal per message. Define the core function on `&IcuAnalysis` and keep the `&IcuMessage` signature as a thin convenience wrapper. Relevant once catalogs reach the 10k+ range (sebastian-software/palamedes#154).

**2. User-facing kind names instead of Debug formatting.**
The diagnostic texts render the kind with `{:?}` ("unsupported formatter kind `List`"). Palamedes plugins surface `diagnostic.message` verbatim to developers, where the ICU keyword (`list`, `number`, `date`) reads better than the Rust enum variant. A `Display` impl on `IcuArgumentKind` fixes this everywhere at once.

**3. Document the kind-vs-style decision; two regression tests.**
- `UnsupportedStyle` with `style: None` renders "style `<none>`" — robust, but the docs should state when a consumer is expected to return `UnsupportedStyle` vs `UnsupportedKind` (a missing style is closer to a default-style/kind decision).
- Add a test with a mixed report (several formatters, different decisions in one message).
- Add a test pinning down that plural/select arguments do **not** flow through the formatter callback — that boundary deserves a regression guard.

swernerx:
Closes #46.

## Summary
- Adds `validate_icu_formatter_support_from_analysis(&IcuAnalysis, ...)` for callers that already ran `analyze_icu` and want to avoid a second AST walk.
- Keeps `validate_icu_formatter_support(&IcuMessage, ...)` as the existing convenience wrapper.
- Implements `Display` for `IcuArgumentKind` and uses user-facing formatter kind names in support diagnostics.
- Documents the `UnsupportedKind` vs `UnsupportedStyle` decision, including missing style/default-style handling.
- Adds regression coverage for mixed formatter decisions and for plural/select arguments not being passed to the formatter callback.

## User-visible impact
- New public ICU API exported by both `ferrocat-icu` and the `ferrocat` umbrella crate.
- Formatter-support diagnostic messages now use ICU-style lowercase kind names such as `list` and `number` instead of Rust debug variant names.

## Validation
- `cargo test -p ferrocat-icu --all-features`
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps`

### DIFF
--- a/crates/ferrocat-icu/README.md
+++ b/crates/ferrocat-icu/README.md
@@ -17,6 +17,7 @@ Use this crate when you want the ICU-specific surface directly:
 - `analyze_icu` for structured argument, formatter, plural, select, and tag summaries
 - `compare_icu_messages` for source/translation compatibility diagnostics
 - `validate_icu_formatter_support` for mapping consumer runtime support policies to diagnostics
+- `validate_icu_formatter_support_from_analysis` when the caller already has an `IcuAnalysis` and wants to avoid a second message walk
 - `normalize_message_metadata` for progressive source-side metadata around `msgid + msgctxt`
 - `extract_argument_names` and `extract_tag_names` when tags should not be mixed with data arguments
 - `extract_variables`, `has_plural`, `has_select`, and related helpers for AST inspection
--- a/crates/ferrocat-icu/src/analysis.rs
+++ b/crates/ferrocat-icu/src/analysis.rs
@@ -1,4 +1,7 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt,
+};
 
 use crate::ast::{IcuMessage, IcuNode, IcuOption, IcuPluralKind};
 
@@ -38,6 +41,24 @@ pub enum IcuArgumentKind {
     Plural,
     /// Ordinal plural expression.
     SelectOrdinal,
+}
+
+impl fmt::Display for IcuArgumentKind {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter.write_str(match self {
+            Self::Argument => "argument",
+            Self::Number => "number",
+            Self::Date => "date",
+            Self::Time => "time",
+            Self::List => "list",
+            Self::Duration => "duration",
+            Self::Ago => "ago",
+            Self::Name => "name",
+            Self::Select => "select",
+            Self::Plural => "plural",
+            Self::SelectOrdinal => "selectordinal",
+        })
+    }
 }
 
 /// Classification of an optional formatter style segment.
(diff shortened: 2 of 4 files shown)


swernerx:
Reviewed — the change itself is good, but I'm **not merging** because of the red Codecov checks. Treat this as approve-pending-coverage, not a design objection.

**What's good:**
- Clean refactor: `validate_icu_formatter_support` now delegates to the new `validate_icu_formatter_support_from_analysis(&IcuAnalysis, ...)`, so there's no duplicated AST walk and the precomputed-analysis entry point is a sensible addition for callers that already ran `analyze_icu`. Re-exported from both `ferrocat-icu` and the umbrella crate. ✅
- Switching the diagnostic messages from `{:?}` (Debug) to the new `Display` impl yields proper ICU-style names (`list`, `number`, `selectordinal` as one word) instead of Rust variant names — good user-facing improvement, and the new tests assert the exact message strings.
- Solid coverage of the mixed kind/style decision and the 'plural/select selectors are not passed to the formatter callback' guarantee.

**Why it's red:** `codecov/patch` is **90.53%** (9 uncovered lines, all in `crates/ferrocat-icu/src/analysis.rs`) against a 95.88% target, which also drags `codecov/project` to 95.83% (−0.06%). Those 9 lines are the `Display` match arms for `IcuArgumentKind` that no test ever renders — only `List`, `Number`, and `Date` get exercised via the diagnostic-message tests; the other arms (`argument`, `time`, `duration`, `ago`, `name`, `select`, `plural`, `selectordinal`) are never hit.

**Suggested fix** — one direct `Display` test closes the gap and pushes project coverage back above base (~95.93% > 95.88%):

```rust
#[test]
fn icu_argument_kind_display_covers_all_variants() {
    use crate::IcuArgumentKind::*;
    for (kind, expected) in [
        (Argument, "argument"), (Number, "number"), (Date, "date"),
        (Time, "time"), (List, "list"), (Duration, "duration"),
        (Ago, "ago"), (Name, "name"), (Select, "select"),
        (Plural, "plural"), (SelectOrdinal, "selectordinal"),
    ] {
        assert_eq!(kind.to_string(), expected);
    }
}
```

Push that, and once the checks go green this is good to merge. Per the 'don't change code' instruction I haven't pushed it myself.

PR #80, opened 2026-06-12
fix(po): use durable unique atomic writes

### LINKED ISSUE #48: po: atomic_write lacks fsync and uses a fixed temp file name
## Problem

`crates/ferrocat-po/src/api/file_io.rs:6-19` writes to a temp file in the target directory and renames it — the right basic shape (atomic on POSIX). Two gaps remain:

1. **No `fsync`** on the file (and directory) before the rename. After a crash, some filesystems may expose an empty or truncated target file despite the rename pattern.
2. **Deterministic temp name** (`.{file}.ferrocat.tmp`): two concurrent writers targeting the same file clobber each other's temp contents, so the surviving rename can publish a torn mix.

## Why it matters

`update_catalog_file` is the API teams will wire into CI and build pipelines, where parallel jobs and abrupt termination are normal. Catalog files are the product's source of truth; losing one to a crash undermines the reliability story.

## Proposal

Use `tempfile::NamedTempFile::persist` (unique name, correct cross-platform semantics) and add an `fsync` before persisting for durability. Directory fsync can be optional/platform-gated.

swernerx:
Closes #48.

## Summary
- Replaces the deterministic `.catalog.po.ferrocat.tmp` write path with `tempfile::NamedTempFile` in the target directory.
- Syncs the temp file before persisting it to the target path.
- Syncs the containing directory after persist on Unix platforms.
- Keeps `atomic_write` internal and preserves the public `update_catalog_file` API.
- Adds regression coverage that the legacy fixed temp name is not used or clobbered.

## User-visible impact
- `update_catalog_file` writes are more robust against concurrent writers targeting the same file and more durable across process/system interruption.
- Adds `tempfile` as a normal `ferrocat-po` dependency; `Cargo.lock` is updated.

## Validation
- `cargo test -p ferrocat-po api::file_io::tests`
- `cargo test -p ferrocat-po --all-features`
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -154,6 +154,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "tempfile",
 ]
 
 [[package]]
--- a/crates/ferrocat-po/Cargo.toml
+++ b/crates/ferrocat-po/Cargo.toml
@@ -23,6 +23,7 @@ memchr = "2.7.6"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.145"
 sha2 = "0.11.0"
+tempfile = "3.23.0"
 
 [dev-dependencies]
 ferrocat-conformance = { path = "../ferrocat-conformance" }
--- a/crates/ferrocat-po/src/api/file_io.rs
+++ b/crates/ferrocat-po/src/api/file_io.rs
@@ -1,4 +1,5 @@
 use std::fs;
+use std::io::Write;
 use std::path::Path;
 
 use super::ApiError;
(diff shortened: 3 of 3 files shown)


swernerx:
Reviewed — looks good, merging. ✅

The core change is a clear improvement over the previous fixed `.{file_name}.ferrocat.tmp` path: a unique per-write temp file removes the predictable-name collision/race between concurrent writers, `sync_all` on the file plus the directory fsync on Unix makes the rename durable, and the regression tests (no leftover temp entry; legacy fixed temp name untouched) are good. `atomic_write` stays private and the public `update_catalog_file` API is unchanged.

**Non-blocking follow-up (not a merge blocker):** `tempfile::NamedTempFile` creates the temp file with mode `0600` on Unix, whereas the old `fs::write` path created it under the process umask (typically `0644`). Since `persist` keeps the temp file's permissions, overwriting an existing catalog via `update_catalog_file` will now also narrow its mode to `0600` (group/other lose read access). For `.po`/NDJSON catalogs that other tools or build steps read, it may be worth preserving the target's existing permissions (or applying the umask explicitly) in a follow-up. Worth a quick verify on your target platforms.

PR #77, opened 2026-06-12
chore: release main

swernerx:
:robot: I have created a release *beep* *boop*
---

<details><summary>ferrocat: 1.0.0</summary>

### ⚠ BREAKING CHANGES

* **api:** collapse redundant option fields into CatalogMode ([#102](https://github.com/sebastian-software/ferrocat/issues/102))

### Features

* **api:** add serializable schema outputs ([#107](https://github.com/sebastian-software/ferrocat/issues/107)) ([8e8bff0](https://github.com/sebastian-software/ferrocat/commit/8e8bff054ac76c6ce0aa76d5d7a707c92f24d9d8))
* **api:** add umbrella namespace modules ([#104](https://github.com/sebastian-software/ferrocat/issues/104)) ([f72d1a5](https://github.com/sebastian-software/ferrocat/commit/f72d1a573cff29cb1bc955359c86531355610ca0))
* **api:** collapse redundant option fields into CatalogMode ([#102](https://github.com/sebastian-software/ferrocat/issues/102)) ([2687df6](https://github.com/sebastian-software/ferrocat/commit/2687df6363755b0ef863594168da8be027d34614))
* **features:** add lean parser profiles ([#106](https://github.com/sebastian-software/ferrocat/issues/106)) ([2887beb](https://github.com/sebastian-software/ferrocat/commit/2887bebfedabb7664f15c12912ea96626b0d103b))
* **icu:** expose stringify_icu ([#96](https://github.com/sebastian-software/ferrocat/issues/96)) ([a581dcb](https://github.com/sebastian-software/ferrocat/commit/a581dcb171a1f1321640f9f22f6da165d23c7f06))
* **icu:** validate formatter support from analysis ([#79](https://github.com/sebastian-software/ferrocat/issues/79)) ([7089df6](https://github.com/sebastian-software/ferrocat/commit/7089df67cdb4803fa72dd79e9f077565aefc91cc))

### Bug Fixes

* **api:** re-export CatalogSemantics from the umbrella crate ([a16e699](https://github.com/sebastian-software/ferrocat/commit/a16e6993b09e86d19712edc601b2fc0271ffcdaf)), closes [#47](https://github.com/sebastian-software/ferrocat/issues/47)

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * ferrocat-icu bumped from 0.13.0 to 1.0.0
    * ferrocat-po bumped from 0.13.0 to 1.0.0
</details>

[...]

PR #114, opened 2026-06-19
docs(msrv): document bump policy

### LINKED ISSUE #69: ci: consider running workspace tests under MSRV
## Problem

Two related issues:

1. **The MSRV job runs `cargo check`, not `cargo test`** (`.github/workflows/ci.yml:107`). Tests are type-checked but never executed under 1.93, so MSRV-specific runtime issues (rare but real, e.g. via dependency behavior differences) would slip through. The workspace builds in seconds — running tests is cheap.
2. **Policy contradiction.** AGENTS.md commits to "support a Rust stable release roughly 9-12 months behind current stable when practical"; the README says "align with OXC when practical". `rust-version = 1.93.0` (released ~Jan 2026) is ~5 months old as of this audit — younger than the policy promises. Edition 2024 compounds the floor. For large projects with conservative toolchains, a fresh MSRV is a genuine adoption blocker, and right now the project's own two policy statements and its actual MSRV disagree with each other.

## Proposal

- Switch the MSRV job to `cargo test --workspace --all-features --locked`.
- Make a deliberate MSRV decision: either lower it toward the documented 9-12-month window, or rewrite the policy (in both AGENTS.md and README, in identical words) to what the project actually wants (e.g. "tracks OXC's MSRV"), and document the MSRV-bump rules (is a bump a minor or major release pre-1.0?).

swernerx:
Closes #69.

## Summary
- documents the pre-1.0 MSRV bump rule in README, AGENTS, and docs pages
- keeps the existing OXC-aligned MSRV policy wording unchanged
- leaves CI unchanged because PR #113 already moved MSRV coverage to workspace tests

## Validation
- `git diff --check`
- `pnpm build` from `docs/`

## Notes
MSRV bumps before 1.0 are now documented as minor-release changes that must be called out in the changelog; patch releases should not raise MSRV.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,7 +40,7 @@ Examples:
   plus smoke tests. Add an explicit threshold when executable umbrella code
   grows.
 - If you add or change dependencies, make sure `Cargo.lock` is updated and the locked checks still pass before push.
-- Keep the declared MSRV in `Cargo.toml` covered by CI. The MSRV policy is to align with OXC when practical, while avoiding churn from tracking only the newest stable toolchain.
+- Keep the declared MSRV in `Cargo.toml` covered by CI. The MSRV policy is to align with OXC when practical, while avoiding churn from tracking only the newest stable toolchain. Before `1.0`, raising the MSRV is allowed in a minor release when the changelog calls it out; patch releases should not raise the MSRV.
 
 ## Public API Changes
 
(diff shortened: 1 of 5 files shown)


swernerx:
CI status note:

All non-semver checks are green on this docs-only PR. The failing `Semver compatibility (ferrocat-icu)` and `Semver compatibility (ferrocat-po)` jobs are not caused by this PR's changes. They compare the current pre-release `main` API against the published 0.13.0 crates and fail on already-landed breaking/API changes that are covered by the open release PR #77.

PR #77 is green and bumps the release line to 1.0.0. Once that release path is handled, this PR should be rebased/rerun rather than patched for semver.

PR #115, opened 2026-06-19
feat(po): add bytes parser charset guard

### LINKED ISSUE #59: po: parser robustness — charset declarations are ignored and unknown lines are silently dropped
## Problem

**1. Charset handling.** All parsers take `&str`, so input is UTF-8 by construction — fine as a baseline, but nothing ever reads the PO header's `Content-Type: ... charset=...` on input (the only non-fixture hit is the writer hardcoding `charset=utf-8`, `api/catalog.rs:884`). Consequences: a Latin-1 PO file dies in the caller's `String::from_utf8` where ferrocat cannot produce a helpful error ("file declares ISO-8859-1"); a file that happens to decode as UTF-8 but declares `charset=ISO-8859-1` is accepted silently. (BOM stripping and CRLF normalization in `parse_po` are handled correctly today.)

**2. Silent line drops.** `parse.rs:199` / `borrowed.rs:336`: `LineKind::Other => Ok(())` — any unclassifiable line outside a string is ignored without trace. A typo like `msgstr_ "..."` silently loses content instead of failing. There is no strict mode at PO level (the `strict` flag in `ParseCatalogOptions` only affects ICU plural projection).

## Why it matters

Both behaviors violate the project's own stated principle (AGENTS.md): "Prefer explicit options and safe defaults over implicit fallback behavior when silent fallback could hide user mistakes." Real-world PO corpora from TMS exports and translators are exactly where mis-declared encodings and hand-edit typos occur.

## Proposal

- A bytes-based entry point (`parse_po_bytes`) that sniffs the declared charset and returns a clear error (or opt-in transcoding) for non-UTF-8 declarations; an audit check "non-UTF-8 charset declared" for the QA path.
- An opt-in strict parse mode that reports unclassifiable lines as errors or diagnostics instead of dropping them.

swernerx:
Closes #59.

## Summary
- adds `ferrocat_po::parse_po_bytes` as the byte-oriented owned PO parser
- rejects declared non-UTF-8 `Content-Type` charsets before decoding
- reports invalid UTF-8 input as Ferrocat `ParseError`
- documents the new API in crate docs, README, and the API overview

## Validation
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p ferrocat-po --all-features --locked`
- `cargo test -p ferrocat-po --no-default-features --lib --locked`
- `pnpm build` from `docs/`

## Notes
The new parser is intentionally exposed from `ferrocat-po` only in this PR. The umbrella `ferrocat` crate should not re-export newly added lower-crate API until the lower crate version containing it has been published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,13 @@ msgstr "world"
 
 For the common "merge fresh extracted messages into an existing catalog" workflow, `merge_catalog` is the lean Gettext-style entry point. For N-way catalog overlays and `msgcat`-style set operations, use `combine_catalogs`. For release checks across a source catalog and target catalogs, use `audit_catalogs`. For application delivery, compile requested-locale artifacts with fallback and ICU diagnostics. For richer high-level flows across PO and NDJSON storage, the docs site's [API overview](https://sebastian-software.github.io/ferrocat/reference/api-overview) is the best next stop.
 
-`parse_po_borrowed` is the allocation-light PO parser for read-heavy paths. It borrows from the source buffer where possible, but it currently requires LF-only input; normalize CRLF input first or use `parse_po`, which handles line-ending normalization internally.
+`parse_po_bytes` is the byte-oriented owned parser for UTF-8 PO input. It
+rejects declared non-UTF-8 charsets before decoding and reports invalid UTF-8
+through Ferrocat's `ParseError` instead of making callers do their own
+`String::from_utf8` step. `parse_po_borrowed` is the allocation-light PO parser
+for read-heavy paths. It borrows from the source buffer where possible, but it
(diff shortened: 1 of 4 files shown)


swernerx:
CI status note:

The failing semver jobs are the same baseline issue seen on #114: current `main` already contains breaking/API changes relative to the published 0.13.0 crates, and release PR #77 is the green release path that moves the crate line to 1.0.0.

The local release gates for this PR passed, including `cargo package -p ferrocat-icu -p ferrocat-po -p ferrocat -p ferrocat-cli --locked`. After #77 is handled, this PR should be rebased/rerun. The new `parse_po_bytes` API is additive in `ferrocat-po`.

Issue #132, opened 2026-06-22
Move runtime formatter-support diagnostics into artifact compilation


swernerx:
## Context

Palamedes currently adds runtime formatter support diagnostics after calling Ferrocat `compile_catalog_artifact`.

The Palamedes-side code parses final runtime messages again, validates formatter support with `validate_icu_formatter_support`, maps `IcuDiagnosticSeverity` into `CompiledCatalogDiagnostic`, and reconstructs locale/source-key metadata for each compiled ID. This is deeper catalog artifact mechanics than Palamedes should own.

Relevant Palamedes behavior:

- number formatter styles: allow empty style, `percent`, `integer`, `::percent`, `::integer`, and `currency/XXX` skeletons,
- date/time formatter styles: allow empty style plus `short`, `medium`, `long`, `full`,
- unsupported formatter kinds are errors,
- unsupported formatter styles are warnings.

## Proposed API direction

Let Ferrocat artifact compilation accept an optional runtime formatter support policy/callback, for example:

```rust
pub formatter_support: Option<fn(&IcuFormatter) -> IcuFormatterSupport>
```

or an equivalent enum/policy abstraction if a function pointer is not the right public shape.

When provided, Ferrocat should validate each final runtime ICU message and emit `CompiledCatalogDiagnostic` entries directly from `compile_catalog_artifact` / `compile_catalog_artifact_selected`.

## Acceptance criteria

- Ferrocat can emit runtime formatter-support diagnostics during artifact compilation without caller-side report mutation.
- Diagnostics include the same compiled key, source key, and resolved locale metadata as other artifact diagnostics.
- The API works for both full artifact compilation and selected-ID compilation.
- Existing behavior remains unchanged when no formatter support policy is provided.
- Tests cover supported number/date/time styles, unsupported formatter styles, and unsupported formatter kinds.
- Palamedes can remove `push_runtime_icu_formatter_support_diagnostics` and keep only the Palamedes-specific support policy definition, if needed.

swernerx:
Implementation note: Palamedes already has working formatter-support diagnostic code that can be adapted into Ferrocat artifact compilation.

Reference implementation:
- Palamedes PR: https://github.com/sebastian-software/palamedes/pull/212
- Merge commit: https://github.com/sebastian-software/palamedes/commit/71f7a0d853a9960f330bfff38626e3c855f36630
[...]

PR #134, opened 2026-06-22
feat(api): add runtime ICU syntax policy

### LINKED ISSUE #133: Add configurable runtime ICU syntax policy for audit and artifacts
## Context

Palamedes PR https://github.com/sebastian-software/palamedes/pull/212 fixed an audit false positive by post-processing Ferrocat audit diagnostics in Palamedes.

The false positive happens because Ferrocat currently validates catalog strings with strict ICU apostrophe parsing. Palamedes runtime behavior treats ordinary literal apostrophes in messages like `you're` and `We've` as valid by normalizing them before parsing. That means Palamedes now has wrapper logic that:

- re-looks up Ferrocat audit diagnostics by locale/source key,
- reconstructs the message strings Ferrocat audited,
- suppresses `icu.invalid_syntax` when strict parsing fails but the runtime parser view succeeds,
- refreshes audit summary counters after filtering,
- performs similar runtime-ICU alignment for compiled artifact diagnostics.

That is catalog/audit mechanics and should live in Ferrocat behind an explicit option instead of being duplicated in Palamedes.

## Proposed API direction

Add an explicit ICU syntax/parser policy that can be used by catalog audit and compiled artifact generation, for example:

```rust
pub enum IcuSyntaxPolicy {
    Strict,
    RuntimeLiteralApostrophes,
}
```

or an equivalent parser/options abstraction.

The default should remain strict to preserve current Ferrocat behavior.

Apply the selected policy consistently in:

- `CatalogAuditOptions` for source/target `icu_syntax` checks,
- audit ICU compatibility parsing,
- `CompileCatalogArtifactOptions` / `CompileSelectedCatalogArtifactOptions` final runtime ICU validation,
- artifact ICU compatibility parsing.

## Acceptance criteria

- Ferrocat audit accepts messages like `you're available` and `We've got {count, plural, one {...} other {...}}` when the runtime-literal-apostrophe policy is enabled.
- Ferrocat audit still reports real invalid ICU syntax such as `{{name}}`.
- Compiled artifact diagnostics no longer need Palamedes-side post-processing for runtime-valid literal apostrophes.
- ICU compatibility diagnostics are still emitted when a runtime-valid message changes placeholder/selector structure.
- Public docs mention the policy and its default.
- Palamedes can remove its `runtime_icu` helper and audit/artifact diagnostic alignment code after updating to the new Ferrocat API.

swernerx:
## Summary

- add `IcuSyntaxPolicy` to audit and compiled artifact options
- support `RuntimeLiteralApostrophes` by parsing runtime literal apostrophes before strict ICU validation
- expose the policy through `ferrocat-po`, with API docs and tests

Closes #133

## Verification

- `cargo test -p ferrocat-po --all-features --locked api::tests::audit --quiet`
- `cargo test -p ferrocat-po --all-features --locked api::tests::compile --quiet`
- `cargo fmt --all --check`
- full stacked verification also passed on #132 branch: workspace tests, MSRV check, package check, coverage gate, docs build


crates/ferrocat-po/src/api/icu_syntax.rs
```diff
@@ -0,0 +1,19 @@
...
+
+pub(super) fn parse_icu_with_syntax_policy(
+    input: &str,
+    policy: IcuSyntaxPolicy,
+) -> Result<IcuMessage, IcuParseError> {
+    match policy {
+        IcuSyntaxPolicy::Strict => parse_icu(input),
+        IcuSyntaxPolicy::RuntimeLiteralApostrophes => {
+            if input.contains('\'') {
+                parse_icu(&input.replace('\'', "''"))
```
swernerx, line 13: This doubles every apostrophe unconditionally, so an input that's already ICU-escaped (`''`) or uses apostrophe quoting (`'{'`) gets parsed with a different meaning than written. Harmless in practice, since it only feeds validation and the stored message stays untouched, and the policy is explicitly for runtimes that don't do apostrophe quoting. Still worth a one-line doc comment stating that assumption, so nobody later reads this as an escaping bug.

swernerx (commented):
Went over the stacked pair. This one's ready from my side. The backward-compatible split into `_with_icu_options` is clean, and the tests hit the cases that matter: strict still flags literal apostrophes, the runtime policy accepts them, and real syntax errors plus compatibility diagnostics both still come through. One optional doc note inline, nothing blocking.

swernerx:
Addressed the optional review note in `127eef9` by clarifying that `RuntimeLiteralApostrophes` is for runtimes that treat apostrophes as literals, and callers relying on ICU apostrophe quoting should keep `Strict`. CI is green after the update.

PR #135, opened 2026-06-22
feat(api): add artifact formatter support diagnostics

### LINKED ISSUE #132: Move runtime formatter-support diagnostics into artifact compilation

swernerx:
## Summary

- add optional `formatter_support` to `CompileCatalogArtifactOptions`
- attach runtime formatter kind/style diagnostics to compiled artifact diagnostics, including selected-artifact compilation
- expose `IcuFormatterSupportPolicy` through `ferrocat-po` and document the artifact ICU option in the API overview

Stacked on #134. Merge #134 first.

Closes #132

## Verification

- `cargo test -p ferrocat-po --all-features --locked api::tests::compile --quiet`
- `cargo test -p ferrocat-po --all-features --locked api::tests::audit --quiet`
- `cargo fmt --all --check`
- `pnpm install --frozen-lockfile` from `docs/`
- `pnpm build` from `docs/`
- `cargo llvm-cov --workspace --all-features --locked --lcov --output-path target/lcov.info --ignore-filename-regex 'crates/ferrocat-(bench|conformance)/'`
- `cargo llvm-cov report --json --summary-only --output-path target/coverage-summary.json --ignore-filename-regex 'crates/ferrocat-(bench|conformance)/'`


crates/ferrocat-po/src/api/compile_types.rs
```diff
@@ -138,6 +149,33 @@ impl CompileCatalogArtifactIcuOptions {
...
+
+    /// Returns options that validate formatter support with the given callback.
+    #[must_use]
+    pub fn with_formatter_support(mut self, formatter_support: IcuFormatterSupportPolicy) -> Self {
+        self.formatter_support = Some(formatter_support);
+        self
+    }
+}
+
+impl PartialEq for CompileCatalogArtifactIcuOptions {
```
swernerx: This block is what's failing `codecov/patch`, and the reason it's never exercised is that nothing compares `CompileCatalogArtifactIcuOptions` anywhere: no `==`, no `assert_eq!`, and no other struct embeds it (it's only passed by reference). The type is new in #134 and unreleased, so there's no semver cost to dropping `Eq`. And `std::ptr::fn_addr_eq` is best-effort by spec anyway, the compiler may merge or duplicate functions, so equality over a fn pointer isn't something to rely on. My vote: drop `PartialEq, Eq` (remove it from the derive over in #134 and this whole block disappears). If we keep it for consistency with the other options types, it needs an equality test and a doc comment about the fn_addr_eq caveat.

crates/ferrocat-po/src/api/compile.rs
```diff
@@ -642,15 +671,31 @@ fn push_icu_compatibility_diagnostics(
...
+    push_icu_diagnostics_for_compiled_message(
+        report.diagnostics,
+        target.source_key,
+        target.compiled_key,
+        target.locale,
+        artifact,
+    );
+}
+
+fn push_icu_diagnostics_for_compiled_message(
```
swernerx, line 671: Minor consistency point: this passes source_key/compiled_key/locale as three separate args, but #134 just added `CompiledIcuDiagnosticTarget` to bundle exactly those, and both callers already have a `target` in hand. Threading it through would be a bit tidier. Optional.

crates/ferrocat-po/src/api/compile_types.rs
```diff
@@ -44,6 +46,13 @@ pub enum CompiledKeyStrategy {
     FerrocatV1,
 }
 
+/// Callback used to validate runtime support for ICU formatters.
+///
+/// The callback receives each formatter discovered in a final runtime ICU
+/// message and returns whether that runtime supports the formatter kind and
+/// style.
+pub type IcuFormatterSupportPolicy = fn(&IcuFormatter) -> IcuFormatterSupport;
```
swernerx, line 57: Flagging for later, not for this PR: a plain fn pointer means a host can't capture config here, like an allowed-currency set read from settings. That's the deliberate cost of keeping the type `Copy`, so likely fine. Just noting that moving to a trait or boxed `Fn` later would break Copy (and the Eq above).

swernerx (commented):
The formatter-support callback is a good shape, and pulling out `push_icu_diagnostics_for_compiled_message` reads well. One thing blocks merge for me, and it's the same thing failing CI: the patch-coverage miss traces straight to the manual `PartialEq`/`Eq` impl, which also turns out to be unused. Details on the lines.

swernerx:
Addressed the blocking review feedback in `c5fea3c`: removed the unused manual `PartialEq`/`Eq` implementation for `CompileCatalogArtifactIcuOptions`, documented the non-capturing function pointer tradeoff, and threaded `CompiledIcuDiagnosticTarget` through the shared diagnostic helper. Local checks passed: `cargo fmt --all --check`, `cargo test -p ferrocat-po --all-features --locked api::tests::compile`, and `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`. CI and Codecov are green now.

PR #141, opened 2026-06-23
feat(po): add catalog coverage report API

### LINKED ISSUE #137: Add catalog completeness and coverage report API
## Context

Palamedes currently implements its completeness report in `packages/cli/src/commands/report.ts`. The report walks source and target PO catalogs and computes per-locale counters:

- total source messages
- translated messages
- missing messages
- fuzzy messages
- completion percent

Ferrocat already audits missing translations and fuzzy flags, but it does not expose a first-class summary report for coverage/completeness. Callers that want a dashboard or CI threshold currently need to rebuild this counting logic themselves.

Completeness and coverage are host-neutral catalog analytics. They are not tied to Palamedes config, Vite, Next.js, or macro extraction. The semantics for what counts as active, obsolete, fuzzy, missing, and translated should be consistent with Ferrocat's normalized catalog model and audit behavior.

## Proposed scope

Add a `ferrocat-po` API that produces a structured coverage report from normalized catalogs, likely keyed by locale.

Possible output shape:

- source message count
- target locale count
- per-locale total/translated/missing/fuzzy/obsolete/extra counts
- per-locale completion percentage
- optional list of missing/fuzzy message identities for callers that want details

The API should operate on already parsed/normalized catalogs. File discovery, config loading, CLI table formatting, JSON printing, and threshold errors should remain caller responsibilities.

## Acceptance criteria

- Source locale entries define the expected active message set.
- Obsolete entries do not satisfy completeness.
- Fuzzy entries are counted separately and do not count as translated by default.
- Empty translations are counted as missing.
- The report supports multiple target locales in one call.
- Palamedes can replace the counting logic in its CLI report with this API while keeping its own config and presentation layer.

## Non-goals

- Do not add a Ferrocat CLI command in this issue unless the API shape needs a small example.
- Do not include Palamedes-specific `pseudoLocale` filtering; callers can decide which locales to pass.

swernerx:
Closes #137.

## Summary
- add a public catalog coverage report API with per-locale totals, incomplete counts, ratios, and optional message details
- share message-status classification with audit so translated, fuzzy, missing, empty, obsolete, and extra statuses are consistent
- re-export the new API through `ferrocat-po` and `ferrocat`, and document it in README and the API overview

## Verification
- `cargo test -p ferrocat-po --all-features --locked coverage`
- `cargo test -p ferrocat-po --all-features --locked api::tests::audit`
- `cargo fmt --all --check`


crates/ferrocat-po/src/api/coverage.rs
```diff
@@ -0,0 +1,358 @@
...
+                    status: CatalogMessageStatus::Extra,
+                });
+            }
+        }
+    }
+
+    coverage
+}
+
+fn increment_status(coverage: &mut CatalogLocaleCoverage, status: CatalogMessageStatus) {
```
swernerx, line 204: This `Extra` arm can't be reached through the real flow: `classify_expected_message` never returns `Extra`, and the extra-detection loop bumps `coverage.extra` directly instead of going through here. So the arm exists only to be exhaustive, and `increment_status_counts_extra_status` exists only to cover it. Same shape turned up in #135 (unused PartialEq) and again in #142 (the `Invalid` MT status). Cleanest fix: route the extra loop through `increment_status(Extra)` too, so the arm is real and the synthetic test goes away. Otherwise drop `Extra` from this match.

crates/ferrocat-po/src/api/coverage.rs
```diff
@@ -0,0 +1,358 @@
...
+            mode: CatalogMode::IcuPo,
+            ..ParseCatalogOptions::new(content, "en")
+        })
+        .expect("parse catalog")
+        .into_normalized_view()
+        .expect("normalize catalog")
+    }
+
+    #[test]
+    fn catalog_coverage_counts_locale_statuses() {
```
swernerx: This inline test (and `catalog_coverage_filters_requested_locales` below it) repeats what tests/coverage.rs already covers through the public API. Inline tests earn their spot when they reach a private fn, like `increment_status_counts_extra_status` does; the duplicated public-path ones could just live in tests/coverage.rs.

swernerx (commented):
Read the stacked set in order, starting here. Strong piece of work, and the thing I flagged on the issue is handled exactly right: the shared `message_status` module is now the single source of truth, and audit was refactored onto it instead of keeping its own copy of the missing/fuzzy/empty rules. Counters, completion ratio, and locale filtering all read cleanly. Two cleanup notes inline, neither blocking.

crates/ferrocat-po/src/api/coverage.rs
swernerx: Following up to make this an explicit call rather than leave it open: the `increment_status_counts_extra_status` test should go. A test that reaches into a private fn purely to tick an otherwise-unreachable match arm doesn't assert any real behavior, it only games the coverage number, which is worse than the gap it fills.

Preferred resolution, in order:

1. Make the arm genuinely reachable by routing the extra-detection loop through `increment_status(Extra)` instead of bumping `coverage.extra` directly. Then the real tests already cover it and nothing synthetic is needed.
2. If you'd rather keep the two counting paths separate, drop `Extra` from this match, mark the path unreachable, and exclude that one spot from coverage (`#[coverage(off)]` on the helper, or an `llvm-cov` region/line exclusion) so the report stays honest.

Either way, remove the synthetic test. Carving a known-unreachable slice out of coverage is the right move here; asserting something fake to keep the percentage up is not.

PR #142, opened 2026-06-23
feat(po): add catalog review report API

### LINKED ISSUE #139: Add catalog diff and translator review report API
## Context

The Palamedes/Ferrocat boundary review did not find another direct Palamedes function that should simply move over. The bigger missing capability is a host-neutral way to summarize catalog changes for translator handoffs and review workflows.

Ferrocat already has parsing, normalized catalogs, audit diagnostics, combine/merge behavior, obsolete handling, fuzzy flag visibility, and machine-translation metadata hashing. What is missing is a structured diff/review report that ties those primitives together.

Catalog diff semantics are format and catalog-model concerns, not framework concerns. Callers should not need to manually compare PO or NDJSON records to answer common workflow questions such as:

- which source messages were added, removed, or changed;
- which target translations are now stale or missing;
- which entries are fuzzy, obsolete, or target-only;
- which machine-translation metadata is valid, stale, or absent;
- which changes need translator attention.

This would help Palamedes and other consumers build review dashboards or CI reports without inventing slightly different catalog comparison rules.

## Proposed scope

Add a read-only `ferrocat-po` API that compares two catalog states or a source/target catalog set and returns a structured review report.

Possible report sections:

- source additions/removals/identity changes;
- target missing/extra/obsolete/fuzzy entries;
- changed translations by identity;
- machine-translation metadata validity/staleness where available;
- summary counters suitable for CI and dashboards.

The API should work from normalized catalogs. File discovery, Git integration, PR comments, and CLI presentation should remain outside the core API.

## Acceptance criteria

- The report is deterministic and keyed by canonical catalog identity (`msgctxt` + `msgid`).
- Obsolete and fuzzy semantics match existing audit behavior.
- Machine-translation metadata checks reuse Ferrocat's existing hash validation rules.
- The API can produce summary-only output without forcing callers to keep large detail lists.
- The API is independent of Palamedes and does not assume a specific repository layout.

## Non-goals

- Do not reintroduce XLIFF scope as part of this issue.
- Do not add GitHub/PR automation to Ferrocat.
- Do not build a Palamedes-specific reviewer UI here.

swernerx:
Closes #139.

Stacked on #141 / `codex/issue-137-coverage-report`.

## Summary
- add a catalog review report API that compares previous and current catalog sets
- report source additions/removals, per-locale translation changes, and machine-translation freshness
- reuse the coverage API from #141 for locale completeness summaries and document the workflow

## Verification
- `cargo fmt --all --check`
- `cargo test -p ferrocat-po --all-features --locked review`


crates/ferrocat-po/src/api/review.rs
```diff
@@ -0,0 +1,844 @@
...
+                source_key: source_key.clone(),
+                status,
+            });
+        }
+    }
+
+    report
+}
+
+fn machine_translation_status(message: &CatalogMessage) -> CatalogMachineTranslationStatus {
```
swernerx, line 413: `machine_translation_status` only ever returns Absent / Current / Stale. Nothing produces `Invalid`, so the variant, its arm in `increment_machine_translation_status`, and `increment_machine_translation_status_counts_invalid_status` are all dead. Same pattern as the `Extra` arm in #141. Two clean options: actually produce it by running `validate_machine_translation_metadata` and returning `Invalid` on failure (the doc comment already promises invalid-metadata detection, so that's likely the intent), or drop the variant until there's a producer.

crates/ferrocat-po/src/api/review.rs
```diff
@@ -0,0 +1,844 @@
...
+    );
+    let coverage = catalog_coverage(
+        current_catalogs,
+        &CatalogCoverageOptions {
+            source_locale: options.source_locale,
+            locales: options.locales,
+            include_details: options.include_details,
+        },
+    )
+    .expect("validated current catalogs must produce a coverage report");
```
swernerx: This `.expect()` leans on catalog_review's validation being a strict superset of catalog_coverage's. True today, but it's an invisible coupling: if coverage ever grows a validation branch, this panics instead of returning an error. Propagating with `?` is safer and reads about the same.

crates/ferrocat-po/src/api/review.rs
```diff
@@ -0,0 +1,844 @@
...
+    current_source_keys: &BTreeSet<CatalogMessageKey>,
+    include_details: bool,
+) -> CatalogTranslationChangeReport {
+    let Some(previous_target) = previous_target else {
+        return CatalogTranslationChangeReport::default();
+    };
+    let mut report = CatalogTranslationChangeReport::default();
+
+    for source_key in current_source_keys {
+        if classify_expected_message(current_target, source_key) != CatalogMessageStatus::Translated
```
swernerx, line 355: Worth documenting: translation changes are tracked only when the current status is `Translated`. A translation that went translated -> empty, or picked up a fuzzy flag, won't show up as a change here; it only surfaces in the coverage counters. Defensible scope, but someone reading 'changed translations' might expect emptied/un-fuzzied entries too. Either note the boundary in the doc or widen it.

swernerx (commented):
Builds on #141 cleanly, and both points I raised on the issue landed: it reuses `catalog_coverage` and the shared status taxonomy instead of re-counting, and rename detection is explicitly out of scope (removal + addition, documented in the API doc). A few things inline, one of them the same dead-variant pattern as #141.

crates/ferrocat-po/src/api/review.rs
```diff
@@ -0,0 +1,844 @@
...
+fn owned_translation(value: EffectiveTranslationRef<'_>) -> CatalogReviewTranslation {
+    match value {
+        EffectiveTranslationRef::Singular(value) => {
+            CatalogReviewTranslation::Singular(value.to_owned())
+        }
+        EffectiveTranslationRef::Plural(values) => CatalogReviewTranslation::Plural(values.clone()),
+    }
+}
+
+fn index_catalogs<'a>(
```
swernerx: `index_catalogs` and `select_target_locales` here are near-identical to the ones in coverage.rs (this copy just takes a `label`). Since coverage and review are meant to share a foundation anyway, these two are the obvious next thing to lift into a shared spot rather than maintain twice.

crates/ferrocat-po/src/api/review.rs
swernerx: Two explicit asks here so this can be picked up later instead of staying ambiguous.

First, same line as #141: the `increment_machine_translation_status_counts_invalid_status` test should be removed. It drives a private fn with a status the classifier never returns, so it asserts nothing real and only props up the coverage number.

Second, and this is why it's different from #141's dead arm: `Invalid` isn't defensively-unreachable, it's an unimplemented feature. The doc comment promises "metadata is invalid in a way the parsed catalog representation can expose", but `machine_translation_status` never actually checks. So please don't paper over it with a coverage exclusion. Pick one:

- wire it up now: run `validate_machine_translation_metadata` (or whatever the parsed representation exposes) and return `Invalid` on failure, then cover it with a real invalid-metadata fixture; or
- pull the `Invalid` variant and its counter out of this PR and track the detection as a follow-up issue, so the public enum doesn't ship a state nothing can produce.

The thing to avoid is the middle ground we have now: a documented status with no producer, plus a synthetic test standing in for the missing logic.

PR #143, opened 2026-06-23
feat(icu): add ICU-aware pseudolocalization

### LINKED ISSUE #136: Add ICU-aware pseudolocalization for catalog artifacts
## Context

While reviewing Palamedes for catalog logic that should live in Ferrocat, the remaining clear gap was pseudo-localization. Palamedes currently has a local `pseudolocalize_message` helper in `crates/palamedes/src/catalog_artifact/compile.rs` that maps characters directly after artifact compilation.

That is useful for UI/layout testing, but the current shape is too naive for a catalog foundation: it rewrites every character, so an implementation like that can mutate ICU argument names, formatter keywords/styles, selectors, and rich-text tag names.

Pseudo-localization is not Palamedes-specific. It is a general catalog QA operation, and Ferrocat already owns the ICU parser/AST plus runtime artifact compilation. A safe implementation should understand ICU message structure and only pseudo-localize literal text while preserving the message contract.

This would let Palamedes delegate pseudo-locale artifact generation instead of carrying a local string-level transformation.

## Proposed scope

- Add an ICU-aware pseudo-localization API in `ferrocat-icu` or the catalog API layer.
- Preserve placeholders, argument names, selector keys, formatter kinds/styles, plural offsets, pound placeholders, and tag names.
- Pseudo-localize only literal text nodes, including nested plural/select/tag branches.
- Provide options for common behavior such as text expansion/wrapping if that can be kept small and stable.
- Expose a catalog/artifact-level path so callers can pseudo-localize final compiled runtime messages without re-parsing outside Ferrocat.

## Acceptance criteria

- `Hello {name}` keeps `{name}` intact and pseudo-localizes only `Hello`.
- Plural/select messages keep selectors and argument structure intact.
- Rich-text tags keep tag names intact while pseudo-localizing child text.
- Invalid ICU input reports diagnostics or errors consistently with the existing parser policy.
- Palamedes can replace its local `pseudolocalize_message` helper with the Ferrocat API.

## Non-goals

- Do not add Palamedes config handling or pseudo-locale routing to Ferrocat.
- Do not render framework-specific modules.

swernerx:
Closes #136.

## Summary
- add ICU-aware pseudolocalization for literal text while preserving arguments, selectors, plural `#`, and tag names
- add compiled artifact pseudolocalization that transforms runtime messages only, preserving missing records and diagnostics
- re-export the API through `ferrocat-icu`, `ferrocat-po`, and `ferrocat`, with README and API overview docs

## Verification
- `cargo test -p ferrocat-icu --all-features --locked pseudolocalize`
- `cargo test -p ferrocat-po --all-features --locked pseudolocalize`
- `cargo fmt --all --check`


crates/ferrocat-po/src/api/pseudolocalization.rs
```diff
@@ -0,0 +1,87 @@
...
+///
+/// Returns [`ApiError::Unsupported`] if any runtime message in the artifact
+/// cannot be parsed as ICU MessageFormat by Ferrocat's parser.
+pub fn pseudolocalize_compiled_catalog_artifact(
+    artifact: &CompiledCatalogArtifact,
+    options: &IcuPseudolocalizationOptions<'_>,
+) -> Result<CompiledCatalogArtifact, ApiError> {
+    let mut pseudolocalized = artifact.clone();
+    for message in pseudolocalized.messages.values_mut() {
+        *message = pseudolocalize_icu(message, options).map_err(|error| {
```
swernerx: This parses every artifact message with strict `parse_icu`. But an artifact compiled under `IcuSyntaxPolicy::RuntimeLiteralApostrophes` (#134) can legitimately contain literal apostrophes like `you're`, which strict parsing rejects, so pseudolocalizing that artifact fails with Unsupported. The two features don't compose right now. I'd thread an `IcuSyntaxPolicy` (or reuse `parse_icu_with_syntax_policy`) through this path so it matches how the artifact was compiled. Minor while you're here: the error doesn't name the offending compiled key, which `iter_mut()` would let you include.

swernerx (commented):
Really clean, and it covers everything I raised on the issue: the AST transform lives in ferrocat-icu, the artifact path in ferrocat-po, there's an explicit re-escape/roundtrip test, and a stability-under-double-application test. Expansion is concretely specified instead of hand-wavy. One integration point worth thinking about, inline.

PR #144, opened 2026-06-23
feat(po): add artifact provenance report API

### LINKED ISSUE #138: Expose optional resolution provenance in compiled catalog artifacts
## Context

Palamedes adds host-facing artifact metadata around Ferrocat's compiled catalog artifact. In `crates/palamedes/src/catalog_artifact/types.rs`, the result includes `resolvedLocaleChain`, and missing-message records already expose `requestedLocale` and `resolvedLocale`.

Ferrocat's artifact compiler already resolves messages through requested locale, fallback locales, and optional source fallback. It records missing messages when the final value did not come from the requested locale. However, consumers do not currently get a complete structured provenance view for all compiled messages.

## Why this may belong in Ferrocat

Resolution provenance is produced by Ferrocat's artifact compiler, not by host adapters. It is useful for build diagnostics, debugging fallback behavior, coverage dashboards, and runtime telemetry. Callers should not need to reconstruct it from `missing` records or duplicate the fallback resolution loop.

## Proposed scope

Add an optional provenance/reporting surface for compiled catalog artifacts.

Possible shape:

- keep the existing artifact wire format stable unless a schema bump is intentional;
- expose an opt-in compile option or side report rather than adding noisy metadata by default;
- for each compiled key, record the source identity, requested locale, resolved locale, and resolution kind (`requested`, `fallback`, `sourceFallback`, `unresolved`);
- optionally include the configured fallback chain used for compilation.

## Acceptance criteria

- Messages resolved from the requested locale are distinguishable from fallback messages.
- Messages resolved from source fallback are distinguishable from configured fallback locales.
- Unresolved messages remain represented consistently with existing `missing` behavior.
- The feature works for both full artifact compilation and selected-ID artifact compilation.
- Palamedes can use the Ferrocat-provided provenance instead of maintaining its own artifact-level fallback metadata.

## Open questions

- Should provenance be embedded in `CompiledCatalogArtifact`, returned as a sibling report, or exposed through a separate `compile_*_with_provenance` API?
- If embedded, does this require a compiled artifact schema version bump?

swernerx:
Closes #138.

## Summary
- add a report-oriented artifact compile API that returns the unchanged `CompiledCatalogArtifact` plus a sibling provenance report
- support both full catalog compilation and selected compiled IDs through one options type instead of another function cross-product
- report requested, fallback, source_fallback, and unresolved resolution kinds with key, source key, requested locale, and resolved locale metadata

## Verification
- `cargo test -p ferrocat-po --all-features --locked compile_catalog_artifact_report`
- `cargo test -p ferrocat-po --all-features --locked compile_catalog_artifact_selected`
- `cargo test -p ferrocat-po --all-features --locked compile_option_constructors`
- `cargo fmt --all --check`


crates/ferrocat-po/src/api/compile_types.rs
```diff
@@ -471,6 +526,66 @@ pub struct CompiledCatalogArtifact {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(deny_unknown_fields))]
+pub struct CompiledCatalogResolution {
+    /// Stable runtime key derived from the source identity.
+    pub key: String,
+    /// Original gettext identity preserved for diagnostics and tooling.
+    pub source_key: CatalogMessageKey,
+    /// Requested locale for this artifact compilation.
+    pub requested_locale: String,
```
swernerx: `requested_locale` is already on `CompiledCatalogProvenanceReport`, and it's identical for every row (it comes straight from `options.requested_locale`). Repeating it on each `CompiledCatalogResolution` clones the same string per message and leaves room for a row to disagree with the header. I'd drop it from the row and keep it only at the report level. `resolved_locale` + `kind` already carry the per-row signal.

crates/ferrocat-po/src/api/compile.rs
```diff
@@ -531,11 +602,61 @@ fn compile_catalog_artifact_from_source_keys<I>(
...
+            .expect("provenance report requested for artifact compile output"),
+    })
+}
+
+fn compile_catalog_artifact_from_source_keys_inner<I>(
+    locales: &BTreeMap<String, &NormalizedParsedCatalog>,
+    source_keys: I,
+    options: &CompileCatalogArtifactOptions<'_>,
+    icu_options: &CompileCatalogArtifactIcuOptions,
+    include_provenance: bool,
```
swernerx: The `include_provenance: bool` + `Option<...>` + `.expect()` on the report path is a small smell: a bool arg that silently decides whether a later `.expect()` holds. It works, but encoding presence in the return type (or just always building the provenance vec and dropping it on the non-report path, if the allocation is negligible) would remove the expect. Optional.

swernerx (commented):
This is exactly the shape I argued for on the issue, and it's nicely done: one `compile_catalog_artifact_report` entry point with a `selection` enum covering both full and selected, ICU options folded in, so provenance doesn't fan out a new cross-product of compile functions. Sibling report, no schema bump, and a test pinning the artifact wire shape unchanged. Two notes inline, both optional.

PR #148, opened 2026-06-23
ci(fuzz): add scheduled parser fuzzing

### LINKED ISSUE #145: ci: add scheduled fuzzing for parser and catalog inputs
## Context

Technical review of `origin/main` at `638bd2b` found that fuzz harnesses already exist, but are intentionally local-only.

Evidence:

- `fuzz/README.md` says: "The first target set is intentionally local-only. Scheduled CI fuzzing should be added after these harnesses have produced stable, useful corpora."
- Existing targets cover the right untrusted-input surfaces:
  - `parse_po`
  - `parse_po_borrowed`
  - `parse_catalog_po`
  - `parse_catalog_ndjson`
  - `parse_icu`
  - `merge_catalog`
  - `update_catalog`
- The normal gates are strong (`fmt`, Clippy, tests, rustdoc, coverage, conformance), but they are finite examples. Parser/catalog fuzzing is the missing resilience loop for malformed or adversarial catalog content.

## Why this matters

Ferrocat parses PO, NDJSON, and ICU MessageFormat input that can come from translators, automation, external systems, or host framework tooling. Parser panics, pathological runtime, and memory blowups are exactly the class of bugs fuzzing is good at catching before releases.

## Suggested work

- Seed initial corpora from `conformance/fixtures`, generated benchmark fixtures, and a few minimal invalid inputs.
- Add a scheduled and manually runnable GitHub Actions workflow for bounded `cargo fuzz run` smoke jobs.
- Start with short per-target budgets so this is a stability signal, not an expensive gate.
- Upload crash artifacts and minimized reproducers when a target fails.
- Add any discovered crash inputs back as regression tests or corpus seeds.
- Consider OSS-Fuzz or longer nightly fuzzing later once local corpora stabilize.

## Acceptance criteria

- A CI workflow runs the core fuzz targets on a schedule and via `workflow_dispatch`.
- Crashes, panics, OOMs, and timeouts fail the workflow.
- Fuzzing docs explain local reproduction, corpus location, and how to promote a found input into a regression test.

swernerx:
Closes #145

## Summary

- adds a scheduled and manually runnable fuzz workflow for all existing parser/catalog fuzz targets
- seeds small deterministic PO, NDJSON, and ICU corpora for bounded CI fuzz runs
- documents local cargo-fuzz setup, CI behavior, and how to promote minimized crash inputs
- refreshes the fuzz lockfile to the current 1.2.0 workspace crate versions

## Validation

- `cargo fmt --all --check`
- `cargo check --manifest-path fuzz/Cargo.toml --bins --offline`

`cargo-fuzz` is not installed in my local environment, so I validated that all fuzz targets compile through the fuzz manifest instead of running live libFuzzer jobs locally.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,78 @@
+name: Fuzz
+
+on:
+  workflow_dispatch:
+    inputs:
+      max_total_time:
+        description: "Per-target fuzzing budget in seconds"
+        required: false
+        default: "45"
+  schedule:
+    - cron: "17 3 * * 1"
+
+permissions:
+  contents: read
+
+env:
+  CARGO_TERM_COLOR: always
+  FUZZ_MAX_TOTAL_TIME: ${{ github.event.inputs.max_total_time || '45' }}
+
+jobs:
+  fuzz:
+    name: ${{ matrix.target }}
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - target: parse_po
+            seed_dir: seed-corpus/po
+          - target: parse_po_borrowed
+            seed_dir: seed-corpus/po
+          - target: parse_catalog_po
+            seed_dir: seed-corpus/po
+          - target: parse_catalog_ndjson
+            seed_dir: seed-corpus/ndjson
+          - target: parse_icu
+            seed_dir: seed-corpus/icu
+          - target: merge_catalog
+            seed_dir: seed-corpus/po
+          - target: update_catalog
+            seed_dir: seed-corpus/po
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Set up nightly Rust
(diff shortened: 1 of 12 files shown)


swernerx (commented):
Solid hardening PR. The workflow shape is right: matrix per target with fail-fast off, bounded libFuzzer runs, artifacts uploaded only on failure, and the README crash workflow (reproduce -> tmin -> regression fixture) is genuinely useful. Seeds cover valid and malformed shapes for all three formats, and the action versions match what the rest of the repo already pins (checkout@v6, upload-artifact@v7).

One process note, not a code issue: between the PR body (cargo-fuzz wasn't runnable locally) and this workflow having no `pull_request` trigger, nothing actually exercises the pipeline until the first scheduled or dispatched run. Nightly + cargo-fuzz + the per-target invocation is exactly the kind of chain that breaks on a typo you can't catch by reading. Worth firing a manual `workflow_dispatch` right after merge to confirm it runs green, instead of letting Monday's cron be the first real test.

PR #150, opened 2026-06-23
perf(bench): add benchmark regression checks

### LINKED ISSUE #147: perf: add regression baseline and profile ICU-heavy catalog update path
## Context

Technical review of `origin/main` at `638bd2b` found that benchmark reporting is useful but does not protect against performance regressions, and the current Rust-only benchmark identifies an ICU-heavy update hotspot worth profiling.

Evidence:

- `.github/workflows/benchmarks.yml` runs a scheduled Rust benchmark report, but only uploads an artifact and writes a summary.
- `docs/app/routes/performance/benchmarking.mdx` explicitly says hosted-runner reports are trend visibility and noise detection, not a hard PR gate.
- A local release run of `cargo run --release -p ferrocat-bench -- compare rust-scheduled-v1 --out target/benchmark/rust-scheduled-review.json` produced stable, non-noisy samples:
  - `po-update/catalog-icu-heavy/ferrocat`: `51.4 MiB/s`, `220,531 msg/s`, `cv 0.65%`
  - `icu-parse/icu-nested-1000/ferrocat`: `73.9 MiB/s`, `267,308 msg/s`, `cv 0.14%`
  - `po-parse/mixed-10000/ferrocat-owned`: `234.6 MiB/s`, `1,987,543 msg/s`, `cv 0.73%`
  - `po-parse/mixed-10000/ferrocat-borrowed`: `281.0 MiB/s`, `2,380,840 msg/s`, `cv 0.55%`
  - `po-stringify/mixed-10000/ferrocat`: `1255.9 MiB/s`, `10,655,807 msg/s`, `cv 0.30%`

The ICU-heavy catalog update path is the slowest measured Rust-only path in this profile. That may be acceptable, but it should be profiled and tracked because performance is a documented product property.

## Why this matters

The current test and coverage gates can all pass while a hot catalog workflow gets slower. Ferrocat's README and docs make performance claims, so the repo should have a lightweight way to notice meaningful regressions before they become release notes or published benchmark drift.

## Suggested work

- Add a benchmark comparison step that can compare a PR/head report against a stable baseline from `main` or a checked-in reference-host report.
- Use conservative regression bands and the existing noise metrics rather than failing on tiny hosted-runner movement.
- Keep scheduled artifacts informational, but add a PR-visible status or summary for meaningful regressions.
- Profile `po-update/catalog-icu-heavy` specifically to identify whether the main cost is repeated ICU parsing/projection, plural category resolution, allocation/cloning, serialization, or catalog normalization.
- Document the regression budget and the process for intentionally updating baselines.

## Acceptance criteria

- Benchmark reports can be compared against a baseline and flag meaningful regressions.
- The `catalog-icu-heavy` update path has a profile-backed optimization plan or a documented target budget.
- Performance docs explain which benchmark movements are release-blocking, informational, or expected noise.

swernerx:
Closes #147

## Summary

- adds `ferrocat-bench regression-check` to compare existing benchmark JSON reports
- fails only when matching, non-noisy scenarios exceed a configurable median elapsed regression threshold
- updates the benchmark workflow so PR runs compare a base-SHA report against the current head report
- documents the PR regression policy and the `catalog-icu-heavy` update-path budget/profiling plan

## Validation

- `cargo fmt --all --check`
- `cargo test -p ferrocat-bench --locked`
- `cargo clippy -p ferrocat-bench --all-targets --all-features --locked -- -D warnings`
- `cargo run --release -p ferrocat-bench -- compare rust-scheduled-v1 --out target/benchmark/regression-baseline.json`
- `cargo run --release -p ferrocat-bench -- regression-check --baseline target/benchmark/regression-baseline.json --current target/benchmark/regression-current.json --max-regression-percent 20`

Regression-check smoke output:

```text
Benchmark regression check
profile: rust-scheduled-v1
max-regression-percent: 20.00
passed: 8 failed: 0 skipped-noisy: 0 missing-baseline: 0 missing-current: 0
result: PASS
```

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -54,6 +54,7 @@ jobs:
           mkdir -p target/benchmark
           git worktree add target/benchmark/base "${{ github.event.pull_request.base.sha }}"
           trap 'git worktree remove --force target/benchmark/base || true' EXIT
+          mkdir -p target/benchmark/base/target
           cargo run --release --manifest-path target/benchmark/base/Cargo.toml -p ferrocat-bench -- compare "$BENCHMARK_PROFILE" --out "$GITHUB_WORKSPACE/target/benchmark/${BENCHMARK_PROFILE}-baseline-${GITHUB_RUN_ID}.json"
 
       - name: Run Rust-only benchmark profile
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -5,7 +5,9 @@ on:
     paths:
       - ".github/workflows/benchmarks.yml"
       - "benchmark/**"
-      - "crates/**"
+      - "crates/ferrocat-bench/**"
+      - "crates/ferrocat-icu/**"
+      - "crates/ferrocat-po/**"
       - "Cargo.toml"
       - "Cargo.lock"
   workflow_dispatch:
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,6 +1,13 @@
 name: Benchmark Reports
 
 on:
+  pull_request:
+    paths:
+      - ".github/workflows/benchmarks.yml"
+      - "benchmark/**"
+      - "crates/**"
+      - "Cargo.toml"
+      - "Cargo.lock"
   workflow_dispatch:
     inputs:
       profile:
@@ -27,10 +34,13 @@ jobs:
     timeout-minutes: 30
     env:
       BENCHMARK_PROFILE: ${{ github.event.inputs.profile || 'rust-scheduled-v1' }}
+      CARGO_TARGET_DIR: ${{ github.workspace }}/target
 
     steps:
       - name: Check out repository
         uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
       - name: Set up Rust
         uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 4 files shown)


crates/ferrocat-bench/src/compare.rs
```diff
@@ -364,6 +398,246 @@ impl CompareCliOptions {
...
+    let mut skipped_noisy = Vec::new();
+    let mut missing_baseline = Vec::new();
+    let mut missing_current = Vec::new();
+
+    for scenario in &current.scenarios {
+        let Some(baseline_scenario) = baseline_by_id.get(scenario.id.as_str()) else {
+            missing_baseline.push(scenario.id.clone());
+            continue;
+        };
+        let regression = regression_scenario(baseline_scenario, scenario)?;
```
swernerx, line 593: This `?` means a single scenario changing its semantic digest aborts the whole regression check, so nothing gets reported. That's harsh for the legitimate case: a PR that intentionally changes a benchmark fixture or workload hard-fails here with nothing actionable, since the base SHA can't carry the new digest until after merge. Consider treating a digest mismatch like `missing_baseline` does, its own 'skipped: semantics changed' bucket, so the unchanged scenarios still compare and the changed one is surfaced instead of blocking the run.

.github/workflows/benchmarks.yml
```diff
@@ -1,6 +1,13 @@
 name: Benchmark Reports
 
 on:
+  pull_request:
+    paths:
+      - ".github/workflows/benchmarks.yml"
+      - "benchmark/**"
+      - "crates/**"
```
swernerx: Worth weighing the cost here: `crates/**` on `pull_request` means every Rust-touching PR runs two release builds (the base-SHA worktree baseline plus current) and a full bench, up to the 30-minute timeout. On runners noisy enough that you already skip-on-noise and only fail above 20%, the marginal signal per PR is thin against that. Options if it bites: gate behind a label, narrow the paths, or keep the per-PR check but only on a hand-picked set of hot scenarios. Fine as-is if you've weighed it; flagging because CI-minute spikes are easy to not notice until they hurt.

crates/ferrocat-bench/src/compare.rs
```diff
@@ -364,6 +398,246 @@ impl CompareCliOptions {
...
+    passed: Vec<BenchmarkRegressionScenario>,
+    failures: Vec<BenchmarkRegressionScenario>,
+    skipped_noisy: Vec<BenchmarkRegressionScenario>,
+    missing_baseline: Vec<String>,
+    missing_current: Vec<String>,
+}
+
+impl BenchmarkRegressionReport {
+    fn has_failures(&self) -> bool {
+        !self.failures.is_empty() || !self.missing_current.is_empty()
```
swernerx, line 473: `missing_current` counting as a failure is a good call (a benchmark silently vanishing should trip the guard), but it's non-obvious and untested, same for the semantic-digest abort path above. bench is coverage-exempt so nothing forces it, but these are real CI-affecting behaviors. A test for the missing-current => FAIL case especially would be worth it, since it's the one a reader wouldn't expect.

swernerx (commented):
Good regression guard, and the tests are real: pass / fail / noisy all exercised with concrete fixtures. The semantic-digest check and the noisy-skip are the right instincts. A couple of design points inline, plus one on CI cost.

swernerx:
Addressed the review notes in e6cb029:

- semantic digest changes now render as a dedicated `SKIP semantics-changed` bucket instead of aborting the whole regression check
- `missing_current` is covered with an explicit FAIL test, including the no-current-scenarios case
- the benchmark workflow PR path filter is narrowed to the benchmark harness plus directly measured runtime crates

Local validation:
- `cargo fmt --all --check`
- `cargo test -p ferrocat-bench --locked`
- `cargo clippy -p ferrocat-bench --all-targets --all-features --locked -- -D warnings`

GitHub checks are green on the updated head.

PR #155, opened 2026-06-24
perf: parser/merge optimizations and cross-runtime benchmark comparisons

swernerx:
Performance audit follow-up. Each commit is an independent optimization,
measured against the `ferrocat-bench` fixtures (median MiB/s, release build,
Apple Silicon). The PO core parser/serializer were already excellent
(memchr + NEON SIMD + Cow + state reuse); the wins are in ICU, NDJSON, and
the catalog merge layer.

## Optimizations & measured gains

| Commit | Path | Gain |
|--------|------|------|
| `perf(icu)` jump to syntax bytes with memchr | ICU parse, literal-heavy | **+323%** (4.2x) |
| | ICU parse, args / plural / tags / nested / select | **+23% … +56%** |
| | ICU parse, formatters (low literal share) | +9% |
| `perf(api)` reuse NDJSON line buffer | NDJSON parse (1k / 10k) | **+10.4% / +11.0%** |
| `perf(api)` borrow matched message in merge | update-catalog (1k / 10k) | **+7.7% / +7.8%** |
| `perf(api)` build plural profile once | update-catalog (10% plural) | 0–2% |
| `perf(icu)` bulk-copy apostrophe/style literals | ICU parse, formatters | +4.4% |
| `build` fat LTO + 1 codegen unit | po-parse (untouched path) | **+16%** |

The headline change is the ICU parser: its literal loop advanced one decoded
char at a time (`advance_char` + `push(char)`) with a per-byte close-tag scan,
where the sibling `ferrocat-po` crate already jumps to the next significant
byte with memchr and bulk-copies the slice. ICU did not even depend on memchr.

## Honest notes

- **Plural profile once**: algorithmically a loop-invariant allocation hoist, but
  the measured effect is small (the profile is cheap and only built for plural
  messages). Kept as a correctness/clarity cleanup; impact scales with plural share.
- **Apostrophe/style memchr**: workload-dependent; larger for long quoted literals,
  modest for the typically short formatter styles.
- **LTO**: lengthens release builds (~7s → ~36s for the bench crate) and only
  affects this workspace's own binaries — dependency consumers build under their
  own release profile.

## Verification

- `cargo fmt --check`: clean
- `cargo clippy --workspace --all-targets`: 0 warnings (incl. the strict
  `redundant_clone` / `needless_borrow` lints)
- `cargo test --workspace`: 423 passing, 0 failing
- ICU parser correctness additionally checked via 8192 proptest cases,
  the conformance suite, and differential edge cases (non-tag `<`, `#`
  outside plural, multibyte UTF-8 incl. CJK).

## Benchmark comparison targets

The branch also broadens the cross-runtime comparison so the numbers sit on the
optimized base, against maintained and widely-used baselines only (judged by
reverse-dependencies and stars, not download counts):

- **PHP `gettext/gettext`** — parse, stringify, and merge, all digest-identical
  to ferrocat (~7x / ~12x / ~12x slower). Adapter under `benchmark/php`, with a
  workaround for a PoGenerator bug that leaves obsolete multi-line output
  non-round-trippable.
- **`babel.Catalog.update`** (Python) — catalog merge with real semantics.
- Rust `polib` was wired up, then dropped: 0 reverse-deps and ~14 stars don't
  clear the adoption bar.

Measurement fix: the digest-summary build moved outside the timed loop for every
tool, so parse numbers reflect the library rather than bench overhead. Ratios are
unchanged; absolute parse throughput rises across the board.

## Homepage

The homepage now leads with performance: cross-runtime parse/update bar charts
and a craft note on the hand-tuned hot path, with PHP alongside the Node and
Python baselines. Re-measured on the M1 Ultra reference machine.


swernerx:
Blocking review note: the benchmark gate needs fixing before this should merge.

The failing `Rust benchmark report` is not showing an actual ICU parser slowdown. The uploaded artifact has the same semantic digest for `icu-parse/icu-nested-1000/ferrocat`, but Base ran `iterations_per_sample=29` at `110.429 ms` / `72.63 MiB/s`, while Current ran `iterations_per_sample=57` at `142.051 ms` / `110.98 MiB/s`. Throughput improves, but the regression checker compares raw `statistics.median_elapsed_ns`, so it reports `+28.64%` just because the current calibration chose more iterations per sample.

Since this PR changes benchmark measurement boundaries and leaves the required performance gate red, please update the regression check to compare normalized per-iteration/per-unit elapsed time, or force baseline/current to use the same `iterations_per_sample` before enforcing the threshold. Otherwise this gate can fail faster PRs and pass slower PRs depending on calibration.

I also ran `cargo test -p ferrocat-icu -p ferrocat-po -p ferrocat-bench --all-features --locked` locally on the PR head; those focused tests passed.

swernerx:
Good catch, and the exact numbers made this easy to pin down. You're right that the gate compared raw `median_elapsed_ns` while the harness calibrates `iterations_per_sample` per run, so the 29→57 jump alone showed up as +28.6% even though throughput actually went up (72.6 → 111 MiB/s).

Fixed in f530e1b: the check now normalizes to per-iteration time (`median_elapsed_ns / iterations_per_sample`) before applying the threshold. I added two tests — one for this exact calibration case (your icu-nested-1000 numbers now read as a ~35% improvement), and one for the inverse you called out, where a real per-operation slowdown hides behind a smaller iteration count and the old raw comparison would have let it through.

Should go green once CI reruns.

PR #156, opened 2026-06-24
bench: add Rust polib and babel comparison targets

swernerx:
Adds two new external comparison targets to `ferrocat-bench`, picked for being
both widely used and actively maintained (no niche/single-author experiments):

- **Rust `polib`** — the one maintained, adopted Rust PO library; fairest
  same-language in-process peer for parse/stringify (no FFI/subprocess bias).
- **Python `babel`** — the de-facto Python localization library and one of the
  few non-C tools with a real, maintained `Catalog.update`; a meaningful peer
  for the catalog merge path.

## Measured (gettext-*-10000, all with identical semantic digests)

| Path | Comparison | Result |
|---|---|---|
| PO parse | ferrocat vs Rust polib | ferrocat **1.75–1.92×** faster |
| PO stringify | ferrocat vs Rust polib | ferrocat **3.78–4.31×** faster |
| Catalog merge | ferrocat vs babel `Catalog.update` | ferrocat **~25×** faster |
| Catalog merge | ferrocat vs GNU msgmerge | ferrocat **~38–40×** faster (subprocess) |

Every comparison passes the bench's cross-tool correctness digest, so the tools
are confirmed to produce semantically identical output before timing.

## How it works

- `polib-rust-parse` / `polib-rust-stringify` dispatch in-process;
  `is_in_process_implementation` keeps such profiles RustOnly (no Node/Python
  setup). `po_summary_from_polib` maps polib's catalog onto the shared semantic
  summary (polib uses fixed header fields + structured Plural-Forms, so known
  fields are mapped back and Plural-Forms reconstructed via `plural_rules.dump()`).
- `babel` runs its real `Catalog.update(no_fuzzy_matching=True)` to match
  ferrocat's exact-identity merge. babel injects placeholder default headers on
  write, so the persisted artifact's headers are normalized back to the source —
  the same normalization the existing polib merge path already does.
- New profiles: `rust-libraries-v1` (in-process Rust vs Rust) and
  `catalog-update-v1` (ferrocat vs msgmerge vs babel).

## Notes

- `polib` only reads/writes PO (no merge), so it has no catalog-update comparison
  — a genuine capability gap, not an omission.
- The ICU compile-to-runtime axis (`@messageformat/core`, `@formatjs/cli`, Lingui)
  is **intentionally not wired up**: those tools precompile ICU to ASTs, whereas
  `compile_catalog_artifact` resolves catalogs to runtime strings and leaves
  parsing to the client — a direct compile-vs-compile benchmark would compare
  different operations. Documented in `benchmarking.mdx`.

## Verification

- `cargo fmt --check` clean, `cargo clippy --workspace --all-targets` 0 warnings
- existing `rust-scheduled-v1` profile still runs unchanged (dispatch regression check)
- new profiles validated end-to-end across de/fr/pl with matching digests


swernerx:
Folded into #155 (perf branch) so the comparison numbers sit on the optimized base. Also adds a fix there: parse paths now time only the parse and build the digest summary outside the measured loop (the summary-mapping overhead was significant and differed per tool — pulling it out raises ferrocat PO parse ~40%, polib ~24%, with all cross-tool digests still matching). The bench branch is superseded.

PR #159, opened 2026-06-29
ci(release): add linux musl CLI artifact

### LINKED ISSUE #158: Add x86_64 musl validation and CLI release artifact
## Context

Palamedes wants to add linux-x64-musl binaries for hosted and containerized environments, especially Alpine/distroless-style deployments. Before adding Palamedes npm packages for that target, Ferrocat should prove that its Rust crate graph builds and tests cleanly for `x86_64-unknown-linux-musl`.

This issue should also make Ferrocat produce a real built artifact for the same target. That keeps Ferrocat support symmetrical with downstream native targets: the lower crate graph is validated, and the Ferrocat CLI itself can be consumed as a musl-friendly binary.

## Proposed scope

- Add a CI validation path for `x86_64-unknown-linux-musl`.
- Run at least `cargo build` for the workspace on that target.
- Run tests where practical. If full test execution is not viable under musl in GitHub Actions, document the limitation and keep a meaningful compile/package gate.
- Build a release artifact for `ferrocat-cli` targeting `x86_64-unknown-linux-musl`.
- Decide and document the artifact shape, for example a compressed archive containing the `ferrocat` binary plus license/readme metadata.
- Fix or document any dependency blockers that prevent musl support.

## Acceptance criteria

- Ferrocat CI has an explicit musl target gate.
- Release automation produces a `ferrocat` CLI artifact for `x86_64-unknown-linux-musl`.
- The artifact is smoke-tested in CI before upload.
- The supported target list/docs mention whether musl is supported and what is validated.
- Palamedes can use this issue as the upstream readiness signal before adding `linux-x64-musl` native packages.

## Notes

This should stay in Ferrocat first because Palamedes depends on `ferrocat`/`ferrocat-po`; if musl fails in the lower crate graph, publishing a Palamedes musl package would only move the failure downstream.

The Ferrocat CLI artifact does not replace Palamedes artifacts. Palamedes still needs to build its own `core-node-linux-x64-musl` N-API package and `cli-linux-x64-musl` `pmds` binary in the `palamedes` npm scope, because those contain Palamedes-specific code.

swernerx:
## Summary

- adds an explicit `x86_64-unknown-linux-musl` CI job that builds/tests the workspace and smoke-tests the CLI
- adds release automation for a smoke-tested `ferrocat-cli` musl archive plus SHA256 checksum on the CLI GitHub release
- documents the musl artifact shape in README, CLI README, docs guide, project status, and release verification

Closes #158.

## Verification

- [x] `git diff --check`
- [x] YAML parsed with Ruby for `.github/workflows/ci.yml` and `.github/workflows/publish.yml`
- [x] `rustup target add x86_64-unknown-linux-musl`
- [x] `cargo check --workspace --all-targets --all-features --locked --target x86_64-unknown-linux-musl`
- [x] from `docs/`: `pnpm install --frozen-lockfile`
- [x] from `docs/`: `pnpm build`

## Notes

The local host is macOS, so the full musl link/runtime path is covered by the new Ubuntu GitHub Actions job rather than by a local `cargo build --target x86_64-unknown-linux-musl` run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,40 @@ jobs:
       - name: Test workspace with MSRV
         run: cargo +1.93.0 test --workspace --all-features --locked
 
+  musl:
+    name: Linux musl target
+    runs-on: ubuntu-24.04
+    env:
+      MUSL_TARGET: x86_64-unknown-linux-musl
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Install musl linker
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y musl-tools
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ env.MUSL_TARGET }}
+
+      - name: Cache Rust build artifacts
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: musl-${{ env.MUSL_TARGET }}
+
+      - name: Build workspace for musl
+        run: cargo build --workspace --all-targets --all-features --locked --target "$MUSL_TARGET"
+
+      - name: Run workspace tests for musl
+        run: cargo test --workspace --all-features --locked --target "$MUSL_TARGET"
+
+      - name: Smoke test musl CLI
+        run: "./target/${MUSL_TARGET}/debug/ferrocat --help"
+
   test:
     name: ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,7 +45,7 @@ jobs:
 
   publish-cli-musl:
     needs: release-please
-    if: ${{ needs.release-please.outputs.ferrocat_cli_release_created == 'true' }}
+    if: ${{ needs.release-please.outputs.releases_created == 'true' }}
     runs-on: ubuntu-24.04
     permissions:
       contents: write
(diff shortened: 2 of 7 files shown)


swernerx:
## Review — `ci(release): add linux musl CLI artifact`

Nicely scoped change: a dedicated `x86_64-unknown-linux-musl` CI job, a release job that builds/smoke-tests/packages the `ferrocat-cli` musl archive with a SHA256 sidecar, and matching docs. Binary name (`ferrocat`), `LICENSE`/README paths, cache keys, and permissions all check out against the repo. Two things are worth addressing before relying on this in a real release.

### 1. `.sha256` sidecar embeds `target/dist/` and breaks `sha256sum -c` for downloaders — `.github/workflows/publish.yml`

```sh
sha256sum "target/dist/${artifact_name}.tar.gz" > "target/dist/${artifact_name}.tar.gz.sha256"
```

`sha256sum` records the path it was given, so the file content becomes:

```
<hash>  target/dist/ferrocat-<version>-x86_64-unknown-linux-musl.tar.gz
```

A user who downloads the archive and the sidecar into one directory and runs the documented verification (`release-verification.mdx` → "verify its `.sha256` file") gets:

```
sha256sum: target/dist/...tar.gz: No such file or directory
...tar.gz: FAILED open or read
```

The `tar` step already avoids this by using `-C target/dist`; the checksum step should do the same so only the basename is stored:

```sh
( cd target/dist && sha256sum "${artifact_name}.tar.gz" > "${artifact_name}.tar.gz.sha256" )
```

### 2. Verify the path-keyed release-please outputs actually fire under the linked/merged config — `.github/workflows/publish.yml`

The job gates on:

```yaml
if: ${{ needs.release-please.outputs.ferrocat_cli_release_created == 'true' }}
# from: steps.release.outputs['crates/ferrocat-cli--release_created']
```

The `crates/ferrocat-cli--<output>` key format is correct for release-please-action, but this repo combines `cargo-workspace` (`merge: false`) with `linked-versions` (`merge: true`, `groupName: ferrocat`). Under a merged/linked release the per-path `release_created`/`tag_name`/`version` outputs can come back empty, in which case this job is silently skipped and no musl asset is ever uploaded — a no-op that looks green. Worth confirming against an actual (or dry-run) release that these three outputs are populated, since the existing `publish-rust` job relies on the global `releases_created` instead.

### Looks good
- Smoke-test paths (`debug`/`release` `ferrocat`) match the `[[bin]] name = "ferrocat"`.
- `musl-tools` + toolchain `targets` is sufficient for the pure-Rust CLI; musl test binaries run on the glibc host fine.
- Docs across README, CLI README, guide, project-status, and release-verification are consistent with the artifact name and shape.

swernerx:
Addressed the review points in `709de6f`:

- checksum generation now runs from `target/dist`, so the `.sha256` sidecar stores only the archive basename and works with `sha256sum -c` after download
- `publish-cli-musl` now runs on any release creation, resolves the CLI version/tag from path-specific release-please outputs when available, falls back to `cargo pkgid -p ferrocat-cli` + `ferrocat-cli-v<version>`, and explicitly checks that the target GitHub release exists before building/uploading

swernerx:
Re-reviewed — both points are resolved. 👍

1. **`.sha256` sidecar** now runs `( cd target/dist && sha256sum "${ARTIFACT_NAME}.tar.gz" > ... )`, so the file stores only the basename and `sha256sum -c` works for downloaders.

2. **Release-please outputs** — addressed more robustly than I suggested: the job now gates on the global `releases_created`, resolves version/tag with a `cargo pkgid` fallback when the per-path outputs are empty, and adds a `gh release view` check that fails loudly instead of silently skipping the upload. The "green but no artifact" failure mode is gone.

LGTM.

PR #166, opened 2026-06-29
perf(po): cut redundant allocations in catalog import (non-breaking)

swernerx:
## Summary

Non-breaking allocation reductions on the `parse_catalog` path, plus the measurement tooling that quantifies the remaining (breaking) opportunities. This is the "non-breaking first + measure" step agreed after the malloc profiling in #161/#164.

### Fixes (no public type changes)

`parse_catalog` builds an owned `PoFile`, then moves most strings into the catalog representation. Three spots re-allocated unnecessarily:

1. **Header map** — keys/values were cloned into the `BTreeMap`; now moved (the `PoFile` is destructured).
2. **References** — `parse_origin` allocated a fresh `file` string per reference. The new `parse_origin_owned` consumes the owned reference and `truncate`s it in place for the file part — zero new allocation.
3. **Singular msgstr** — was copied via `first_str().to_owned()` even though the owned `String` already existed; now moved out of `MsgStr` via `take_first_msgstr`.

### Results — catalog-modern-de-1000

| metric | before | after | Δ |
|---|---|---|---|
| allocations/item | 7.6 | 5.6 | **−26%** |
| parse-catalog-po throughput | 343 | 400 MiB/s | **+16.5%** |

### Measurement tooling

- `alloc-stats` now also reports `parse_catalog` allocations (next to owned/borrowed).
- New `string-stats` command: string-length buckets and per-item reference/comment counts.

## Data for the next (breaking) step

`string-stats` over the realistic fixtures shows where the remaining ~31% malloc lives and what a breaking representation change would buy:

| | strings ≤24B (CompactString-inline) | refs/item ≤1 | comments/item ≤1 |
|---|---|---|---|
| catalog-modern-de-1000 | 35% | 100% | 100% |
| gettext-ui-de-10000 | 42% | 100% | 99% |

Takeaway for the separate breaking PR: **`SmallVec<[_; 1]>`** on the collection fields would inline the container `Vec` for ~100% of items (the most reliable win), and **`CompactString`** would additionally inline ~35–42% of strings. Both change public `PoItem`/`CatalogMessage` types (semver-major, ripples to Palamedes), so they stay out of this PR.

## Verification

- `cargo test -p ferrocat-po` — green, incl. new unit tests for `parse_origin_owned` and `take_first_msgstr`
- `cargo clippy --workspace --all-targets` (incl. `--features count-alloc`) — clean
- `cargo fmt --all --check` — clean

### DIFF
--- a/crates/ferrocat-bench/src/main.rs
+++ b/crates/ferrocat-bench/src/main.rs
@@ -101,6 +101,11 @@ fn run() -> Result<(), String> {
             let fixture = load_fixture(&fixture_name)?;
             run_alloc_stats(&fixture)
         }
+        "string-stats" => {
+            let fixture_name = args.next().unwrap_or_else(|| "realistic".to_owned());
+            let fixture = load_fixture(&fixture_name)?;
+            run_string_stats(&fixture)
+        }
         "parse-catalog-po" => {
             let fixture_name = args
                 .next()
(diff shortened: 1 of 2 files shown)


swernerx:
Reviewed the PR. I do not see any blocking issues.

The `parse_catalog` allocation reductions look semantically equivalent to the old code path:
- header entries are moved into the `BTreeMap` with the same duplicate-key behavior as the previous cloned iterator;
- `parse_origin_owned` preserves the old `path:line` parsing behavior while reusing the owned reference buffer;
- `take_first_msgstr` matches the previous `first_str().unwrap_or_default()` behavior for singular import without copying.

The measurement additions also look scoped to the bench tooling and the reported fixture numbers line up with a local smoke run. One small non-blocking cleanup for a later pass: the `ferrocat-bench` unknown-command help still omits `alloc-stats`, and now also omits `string-stats`, but that was already incomplete before this PR.

Verified locally on `9a9f5b6`:
- `cargo fmt --all --check`
- `cargo test -p ferrocat-po --locked`
- `cargo check -p ferrocat-bench --features count-alloc --locked`
- `cargo run -p ferrocat-bench --features count-alloc --quiet -- alloc-stats catalog-modern-de-1000`
- `cargo run -p ferrocat-bench --quiet -- string-stats catalog-modern-de-1000`

Current GitHub checks are green, including benchmark report, semver compatibility, packaging, Rustdoc, coverage, and docs build.

PR #167, opened 2026-06-29
feat(po)!: store per-item PO collections inline with SmallVec

swernerx:
## Summary

The per-item PO collections almost always hold a single element — `string-stats` over the realistic fixtures shows **references at 100% ≤1 per item** and comments at 99–100% ≤1 — yet each still forced a heap allocation for its `Vec` backing buffer on every item.

This stores the common single-element case inline by moving `PoItem`'s hot per-item collections from `Vec<T>` to **`PoVec<T>`**, a re-exported `SmallVec<[T; 1]>`:

- `references`, `comments`, `extracted_comments`, `flags`, `metadata` → `PoVec<T>`
- `MsgStr`, `Header`, and the file-level `PoFile` collections keep `Vec`/`String` (not per-item hot paths)

This was scoped down from the originally-discussed CompactString+SmallVec change: **CompactString couldn't be confined to `PoItem`** without either regressing the catalog path (PoString→String reallocation at the boundary) or rewriting the entire public catalog API across 17 files for only a partial benefit (long `msgstr` translations don't inline). SmallVec delivers the biggest, most reliable win cleanly; strings stay `String`, so #166's in-place `parse_origin` reuse and the borrowed path are untouched. CompactString can be revisited as its own effort.

## Results (release, fat LTO)

| metric | before | after |
|---|---|---|
| parse_po allocations/item (catalog-modern) | 4.5 | **3.3** (−27%) |
| parse_catalog allocations/item (catalog-modern) | 5.6 | **4.6** (−18%) |
| parse throughput (catalog-modern) | 527 | **585 MiB/s** (+11%) |
| parse throughput (gettext-ui-10000) | 472 | **518 MiB/s** (+10%) |
| parse-catalog-po throughput | 390 | 385 (flat) |

The clear win is the **parse_po path** (direct `PoFile` consumers): +10% throughput, −27% allocations. The catalog path drops allocations but is throughput-flat — the `PoVec→Vec` conversions at the catalog boundary plus the ICU-heavy per-item work offset the malloc savings there.

### Borrowed parser

The same change is applied to `BorrowedPoItem` (the strings are already borrowed `Cow`, so the container buffers were the only per-item heap allocations):

| metric | before | after |
|---|---|---|
| parse_po_borrowed allocations/item (catalog-modern) | 1.2 | **0.03** |
| parse-borrowed throughput (catalog-modern) | 792 | **915 MiB/s** (+15.5%) |
| parse-borrowed throughput (gettext-ui-10000) | 607 | **703 MiB/s** (+15.8%) |

A single-reference borrowed item is now allocation-free.

### Catalog output types

Carried into the public catalog output so the win reaches **both** catalog formats (PO and NDJSON both produce `CatalogMessage`): `CatalogMessage::origin` → `PoVec<CatalogOrigin>` (origin is the per-message collection that's almost always one element). `comments`/`flags`/`translator_comments` stay `Vec` (usually empty → no allocation; and `merge_unique_strings` is shared with placeholder values). This also removes the `Vec` boundary conversion the earlier commits introduced for references.

| metric | before* | after |
|---|---|---|
| parse_catalog allocations/item (catalog-modern) | 4.6 | **3.6** (5.6 on main) |
| parse-catalog-po | 2493 | **2663 iter/s** (+6.8%) |
| parse-catalog-ndjson | 1783 | 1759 iter/s (flat — bottleneck is serde_json) |

\* before = the owned+borrowed-only state earlier in this PR; bundled here so the catalog output types don't require a second major release.

NDJSON-internal optimizations (the private `NdjsonRecord` etc., and the ~15–18% in serde_json deserialization) are **not** included: the private structs aren't breaking (can ship later as a minor) and replacing serde_json is out of scope.

## Breaking change

`PoItem::{references, comments, extracted_comments, flags, metadata}`, the matching `BorrowedPoItem` fields, and `CatalogMessage::origin` are now `PoVec<T>` (`SmallVec<[T; 1]>`) instead of `Vec<T>`. Read-heavy code is unaffected (Deref to slice, iteration, indexing, serde). Callers that construct a field from a `Vec` now need `.into()`, and direct `PartialEq` against `Vec<_>` needs `.as_slice()` on both sides.

## Version / release note

Versions are **not** bumped here; the breaking commit (`feat(po)!`) drives release-please to bump to **2.0.0** after merge.

Note: `cargo-semver-checks` passes on this PR — it currently has no lint for a public struct **field type** changing (`Vec` → `SmallVec`), so it does not flag this break. The change *is* breaking for code that constructs these fields; the `feat!`/`BREAKING CHANGE` marker (not the semver-checks result) is what guarantees the major bump.

## Verification

- `cargo test --workspace --all-features` — green (incl. property tests, conformance, doctests)
- `cargo clippy --workspace --all-targets --all-features` — clean
- `cargo fmt --all --check` — clean


swernerx:
Review note: I would hold this until the public API facade and docs are updated.

This PR changes public fields from `Vec` to `PoVec` (`CatalogMessage::origin`, plus the low-level `PoItem`/`BorrowedPoItem` per-item collections), but the umbrella `ferrocat` crate still re-exports the affected types without re-exporting `PoVec` or `SmallVec` through `ferrocat::catalog`, `ferrocat::po`, or top level. That means users depending on `ferrocat` alone now see public fields whose type is only conveniently nameable through the lower-level `ferrocat_po` crate. The new alias is explicitly meant to avoid forcing callers to depend on `smallvec` directly, so the facade should mirror that export for the public types it exposes.

There is also a stale API overview example at `docs/app/routes/reference/api-overview.mdx:796`: it still constructs `CatalogMessage { origin: Vec::new(), ... }`, which no longer type-checks because `origin` is now `PoVec<CatalogOrigin>`. Please update the example and the surrounding user-facing API docs to show the new construction pattern (`PoVec::new()`, `Default::default()`, or `.into()`), and consider documenting the breaking Vec-to-PoVec field change in the reference page.

Verified locally: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, `cargo check --workspace --all-targets --all-features --locked`, and `cargo test -p ferrocat-po --all-features --locked`. Live PR checks were green.

swernerx:
Both points addressed in c480af8:

- **Facade re-export.** `PoVec` and `SmallVec` are now re-exported through `ferrocat::catalog`, `ferrocat::po`, and the crate root, so users on `ferrocat` alone can name `CatalogMessage::origin` / the `PoItem`/`BorrowedPoItem` collection fields without reaching for `ferrocat_po` or `smallvec` directly. (At the crate root they're exported via the unconditional `po` group to avoid a duplicate re-export when the `catalog` feature is on.)
- **Stale docs example.** `docs/app/routes/reference/api-overview.mdx` no longer constructs `CatalogMessage { origin: Vec::new(), .. }`; it imports `PoVec`, uses `PoVec::new()`, and a new field-type note documents the breaking `Vec`→`PoVec` change plus how to construct it (`PoVec::new()` / `Default::default()` / `vec![…].into()`).

Verified: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features`, and `cargo test --workspace --all-features` are clean (`ferrocat` builds with both default and all features).

Issue #189, opened 2026-07-01
Performance: targeted micro-optimizations in parser, serializer, and ICU paths


swernerx:
## Summary

A handful of targeted, self-contained performance fixes in the parser, serializer, and ICU layers. Each is a single-site change with no public API impact. (The larger, systemic cloning issues in the `api/` catalog layer are tracked in #188.)

Line numbers refer to the v2.0.0 tree.

### 1. Owned parser clones `msgstr` instead of moving it (plural finalize path)

`crates/ferrocat-po/src/parse.rs:417`:

```rust
let mut values = state.item.msgstr.clone().into_vec();
```

clones the entire msgstr content although the field is overwritten immediately afterwards. Hits every plural item with exactly one msgstr slot, in the owned-parser hot path.

**Fix:** `core::mem::take(&mut state.item.msgstr).into_vec()`.

### 2. `declared_charset` scans the whole file with a naive windowed search

`crates/ferrocat-po/src/parse.rs:220-224` (`find_ascii_case`): a naive O(n·13) `windows().position()` search for `content-type:` over the *entire* input, once per `parse_po_bytes` — ironically one of the few paths that bypasses the advertised memchr scanning.

**Fix:** use `memchr2(b'c', b'C', …)` as the anchor and compare the window only at hits — or bound the search to the first few KB (the header lives at the top of the file).

### 3. ICU parser: `find_literal_stop` scans past the structural stop

`crates/ferrocat-icu/src/parser.rs:537-546`: `memchr3` finds e.g. `{` at offset 5, but the separate `memchr(b'\'')` / `memchr(b'#')` runs keep scanning to their next hit or the end of the remaining input. For messages with many arguments and no apostrophes, the tail is re-scanned per literal run → O(arguments × length).

**Fix:** restrict the follow-up searches to `&haystack[..structural_stop]`.

The same pattern exists in the x86 fallback `fallback_find_escapable_byte` (`crates/ferrocat-po/src/scan.rs:49-56`) — on non-aarch64 targets the serializer escape scan pays triple.

### 4. Serializer: throwaway `String` built just to measure length; intermediate `Vec` per folded value

`crates/ferrocat-po/src/serialize.rs:431-442` (`escaped_part_len`): builds the fully escaped `String`, reads only `.len()`, and discards it. Since `escaped.len() == part.len() + count_of_escapable_bytes`, a counting scan with no allocation suffices.

[...]

PR #197, opened 2026-07-01
perf(po): reduce parser and serializer allocations

swernerx:
## Summary

- Refs #189.
- Moves the owned parser plural-finalization path instead of cloning `MsgStr` before overwriting it.
- Uses a memchr-backed candidate scan for case-insensitive charset lookup.
- Bounds follow-up searches in ICU literal scanning and the PO fallback escape scanner to the first structural stop.
- Avoids serializer throwaway allocations for escaped-length measurement and removes an intermediate multiline-part `Vec`.
- Uses borrowed keys in ICU compatibility maps and sets where diagnostics only need owned strings at emission time.

## Verification

- [x] `cargo test -p ferrocat-icu --all-features --locked`
- [x] `cargo test -p ferrocat-po --all-features --locked`
- [x] `git diff --check`

## Notes

- No public API changes.
- Local validation did not include a full before/after timing run. The PR benchmark workflow should run `rust-scheduled-v1` against the base and current branch; keep #189 open until that report is reviewed.

### DIFF
--- a/crates/ferrocat-icu/src/analysis.rs
+++ b/crates/ferrocat-icu/src/analysis.rs
@@ -466,36 +466,36 @@ fn unique_names<'a>(names: impl IntoIterator<Item = &'a String>) -> Vec<String> 
     let mut seen = BTreeSet::new();
     let mut out = Vec::new();
     for name in names {
-        if seen.insert(name.clone()) {
+        if seen.insert(name.as_str()) {
             out.push(name.clone());
         }
     }
     out
 }
 
-fn argument_map(analysis: &IcuAnalysis) -> BTreeMap<String, IcuArgumentKind> {
+fn argument_map(analysis: &IcuAnalysis) -> BTreeMap<&str, IcuArgumentKind> {
     analysis
         .arguments
         .iter()
-        .map(|argument| (argument.name.clone(), argument.kind))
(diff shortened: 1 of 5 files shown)


swernerx (commented):
Went through this against the five findings in #189 — every one is addressed, and each of the tricky helpers got a dedicated test, which is exactly what I wanted to see here:

- mem::take in the plural finalize path replaces the full MsgStr clone.
- find_ascii_case now anchors on memchr2 over both cases of the first byte instead of the O(n·13) window scan; the empty-needle and non-alpha anchor cases are covered by tests.
- find_literal_stop and the x86 fallback escape scanner both bound their follow-up searches to the first structural stop — the O(arguments × length) rescan is gone.
- escaped_string_len counts instead of building a throwaway String, and the counting logic is verified against escape_string().len() on a mixed input. The intermediate parts Vec in write_complex_keyword is gone too.
- The ICU compatibility maps are keyed by &str now; owned strings only materialize at diagnostic emission, where they're needed anyway.

The borrowed formatter_map comparison (Option<&str> vs the old owned trim) is semantically identical, checked that path. Agree with keeping #189 open until the benchmark report confirms the wins. LGTM — can't formally approve from the author account.

Issue #193, opened 2026-07-01
Feature flags compile/mt/plurals are no-op aliases and the feature set is undocumented


swernerx:
### `compile`, `mt`, `plurals` are no-op aliases

In `crates/ferrocat-po/Cargo.toml` (and mirrored in `crates/ferrocat/Cargo.toml`), the features `compile`, `mt`, and `plurals` all expand to just `["catalog"]`. There is **not a single** `#[cfg(feature = "compile"/"mt"/"plurals")]` in the source — the modules `api/compile.rs`, `api/mt.rs`, and `api/plural.rs` always compile whenever `catalog` is enabled.

The features suggest granular compilation ("lean parser profiles", as advertised in the 1.0.0 changelog) but deliver nothing. Worse, actually gating them later would be a breaking change for existing `features = ["catalog"]` users who currently get those modules for free — so the window to make them real is *now*, while gating is still additive relative to what the alias promises.

### Feature flags are entirely undocumented

Neither README (root, `crates/ferrocat`, `crates/ferrocat-po`) nor any crate-level rustdoc explains the feature set (`full`, `catalog`, `compile`, `mt`, `plurals`, `serde`), what each enables, or what the default is. Users discover them only by reading `Cargo.toml`.

## Proposed fix

1. Decide per alias feature: **gate it for real** (add the `cfg`s to `api/compile.rs`, `api/mt.rs`, `api/plural.rs` and their re-exports — safe now because the features already imply `catalog`) or **remove/mark it** ("reserved, currently implies catalog").
2. Add a feature table to the `ferrocat` and `ferrocat-po` READMEs and crate-level docs, e.g.:

   | Feature | Default | Enables |
   |---|---|---|
   | `catalog` | yes | high-level catalog workflows (parse/update/audit/...) |
   | `compile` | yes | runtime artifact compilation (implies `catalog`) |
   | ... | | |

3. Add `#[cfg_attr(docsrs, doc(cfg(feature = "...")))]` annotations plus the `docsrs` cfg in `[package.metadata.docs.rs]` so docs.rs shows which items need which feature.
4. Cover the matrix in CI (`cargo hack check --each-feature`) — see also #185.

## Severity

Medium — misleading public packaging surface; cheap to fix now, breaking to fix later.

swernerx:
Triage note:

[...]

PR #198, opened 2026-07-01
fix(api): repair umbrella crate re-exports

### LINKED ISSUE #185: Umbrella crate fails to build with --no-default-features
## Problem

The `ferrocat` umbrella crate fails to compile without default features:

```
cargo build -p ferrocat --no-default-features
# error[E0432]: unresolved imports ...
```

In `crates/ferrocat/src/lib.rs`, the `pub mod catalog { pub use ferrocat_po::{CatalogAuditReport, ...} }` module (around line 71) is **not** gated behind `#[cfg(feature = "catalog")]`, while the corresponding root-level re-exports further down (around line 170) are correctly gated. With `--no-default-features`, `ferrocat-po` is built without its `catalog` feature, so the items referenced inside `pub mod catalog` do not exist and the build breaks.

This means the documented optional `catalog` feature is effectively broken in the 2.0.0 release: any consumer using `default-features = false` (e.g. to get a lean parser-only build) cannot compile the crate at all.

## Proposed fix

1. Add `#[cfg(feature = "catalog")]` to the `catalog` module in `crates/ferrocat/src/lib.rs` (and audit the file for any other ungated re-exports of feature-gated items).
2. Add a CI guard so this class of bug cannot ship again:
   - Minimal: a `cargo check -p ferrocat --no-default-features` step.
   - Better: `cargo hack check --workspace --each-feature --no-dev-deps` (via [cargo-hack](https://github.com/taiki-e/cargo-hack)) to validate the full feature matrix, including feature combinations of `ferrocat-po`.

The fix itself is non-breaking and a good candidate for a 2.0.1 patch release.

## Severity

High — released feature configuration does not compile.

### LINKED ISSUE #186: Umbrella crate re-export drift: unreachable API items, duplicated schema constant, bogus deprecation versions
## Problem

The README states that the `ferrocat` umbrella crate "re-exports the stable Rust surface from the lower-level workspace crates", and the crate is positioned as the recommended dependency for application code. In 2.0.0 there are several gaps and drift issues that force users back onto the internal crates — exactly the version coupling the umbrella is supposed to prevent.

### Missing re-exports

- `parse_po_bytes` — advertised in the root README ("byte-oriented ... parsers") and in `ferrocat-po`'s own crate docs, but not reachable via `ferrocat`.
- `ParsePosition` — the return type of `ParseError::position()`. The type is **not nameable** through the umbrella crate, so users cannot write out the type of a value the API hands them.
- `audit_catalogs_with_icu_options` and `CatalogAuditIcuOptions` — public and documented in `ferrocat-po` (`src/lib.rs`), missing from the umbrella.
- The `diagnostic_codes` modules (both crates) — the documented "stable machine-readable codes" for `Diagnostic.code` cannot be referenced through the umbrella.

### Duplicated constant instead of re-export

`crates/ferrocat/src/lib.rs` (~line 163) **re-defines** `COMPILED_CATALOG_ARTIFACT_SCHEMA_VERSION: u16 = 1` instead of re-exporting `ferrocat_po::COMPILED_CATALOG_ARTIFACT_SCHEMA_VERSION` (`crates/ferrocat-po/src/api/compile_types.rs`). The wire-format validation checks against the `ferrocat-po` constant; if that is ever bumped, umbrella users silently compare against a stale copy.

### Incorrect deprecation metadata

`has_selectordinal` and `MergeExtractedMessage` in `crates/ferrocat/src/lib.rs` carry `#[deprecated(since = "0.14.0", ...)]` — a version that never existed (the release history goes 0.13.0 → 1.0.0). These aliases also survived the 2.0 major bump, which would have been the natural point to remove them.

## Proposed fix

1. Complete the re-export list (items above); re-export the schema-version constant instead of duplicating it (under the `catalog` feature gate).
2. Fix the `since` versions on the deprecated aliases (or remove the aliases in the next major; document the plan either way).
3. Add a public-API snapshot test with [`cargo public-api`](https://crates.io/crates/cargo-public-api) (or `cargo semver-checks`) in CI so umbrella/sub-crate drift is caught automatically.
4. Longer term, consider reducing the duplication of canonical paths (every item is currently exported at the root *and* under `po`/`icu`/`catalog` modules), which doubles the drift surface and clutters docs.rs.

Items 1–2 are non-breaking and fit a 2.0.1 patch release.

## Severity

High — documented API surface is unreachable through the recommended crate.

swernerx:
## Summary

- Fixes #185.
- Fixes #186.
- Gates the `ferrocat::catalog` module behind the umbrella `catalog` feature so `cargo check -p ferrocat --no-default-features --locked` compiles again.
- Completes the missing umbrella re-exports for byte parsing, parse positions, audit ICU options, diagnostic code modules, and the compiled artifact schema version.
- Re-exports the schema version from `ferrocat-po` instead of duplicating the constant, and corrects invalid deprecation `since` metadata without removing compatibility aliases.
- Adds CI and smoke-test coverage for the repaired surface.

## Verification

- [x] `cargo check -p ferrocat --no-default-features --locked`
- [x] `cargo test -p ferrocat --all-features --locked`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc -p ferrocat --all-features --no-deps --locked`
- [x] `git diff --check`

## Notes

- Public API impact is additive except for correcting deprecation metadata.
- The broader feature-matrix decision and any `cargo hack --each-feature` adoption stay with #193.


swernerx (commented):
Checked against #185 and #186 — this covers both fully:

- The catalog module is gated, and cargo check -p ferrocat --no-default-features is in CI so the broken configuration can't ship again. Deferring the full feature matrix to #193 makes sense.
- All four re-export gaps are closed (parse_po_bytes, ParsePosition, audit_catalogs_with_icu_options + CatalogAuditIcuOptions, diagnostic_codes in all three modules), and the smoke test actually exercises each of them rather than just naming them.
- The schema version is now derived from the ferrocat-po constant, so a bump can't drift silently — and the smoke test asserts the equality.
- since = "2.0.0" on the deprecation metadata is a reasonable call given 0.14.0 never existed and the aliases survived into 2.0.

One optional follow-up thought, not for this PR: a cargo public-api snapshot in CI would catch the next re-export drift automatically instead of relying on the smoke test growing with the surface. Could live in #193 alongside the feature-matrix work. LGTM — approval blocked because it's the author account.

Issue #188, opened 2026-07-01
Performance: catalog API layer deep-clones every message 3-4x per update_catalog


swernerx:
## Summary

The byte-level core (LineScanner, escape scanning, `Cow`-based borrowed parsing, `merge_catalog`'s borrow-based index) delivers on the zero-copy promise. The real bottleneck for large catalogs sits one layer up, in `crates/ferrocat-po/src/api/`: **a single `update_catalog` run deep-clones every message 3–4 times** (normalize → merge → export), and the origin sort allocates per comparison. For the advertised 10k-message workload this is the dominant avoidable cost.

Line numbers below refer to the v2.0.0 tree.

### 1. `sort_messages` with `OrderBy::Origin` allocates two `String`s per comparison — high impact

`api/catalog.rs:660-673`: the `sort_by` comparator calls `first_origin_sort_key`, which returns `origin.file.clone()` (or `String::new()`) — per comparator invocation, i.e. O(n log n) times. For 10k entries that is roughly 280k heap allocations per `update_catalog` call.

**Fix:** compare borrowed keys (`origins.first().map(|o| o.file.as_str()).unwrap_or("")`) — no allocation needed at all; alternatively `sort_by_cached_key`. Also prefer `sort_unstable_by`: the stable sort allocates a buffer, and stability is already covered by the explicit tiebreakers.

### 2. `update_catalog` deep-clones input it already owns — high impact

`api/catalog.rs:282-344` (`normalize_update_input`) clones `msgid`, `msgctxt`, `comments`, `origin`, and `placeholders` per extracted message; `merge_message` (`catalog.rs:578-584`) then clones the same fields *again* into the `CanonicalMessage`. Yet `update_catalog` takes `options` by value (even carrying `#[expect(needless_pass_by_value)]`) — it owns `options.input` but only passes it on as `&CatalogUpdateInput`.

**Fix:** make `normalize_update_input` consuming (`input: CatalogUpdateInput`), feed `merge_catalogs`/`merge_message` via `into_iter()` over `Vec<NormalizedMessage>`, and move fields instead of cloning.

### 3. Merge index clones keys per entry and per lookup — high impact

[...]

PR #201, opened 2026-07-01
perf(po): avoid catalog sort and dedupe allocations

swernerx:
## Summary

- Refs #188.
- Compares borrowed origin file keys in `OrderBy::Origin` instead of cloning `String`s during sort comparisons.
- Uses `sort_unstable_by` for origin ordering with explicit message/context/obsolete tie-breakers.
- Removes the per-check cloned `BTreeSet` rebuild from helper string dedupe.
- Adds focused unit coverage for borrowed origin-key sorting and origin tie-break behavior.

## Verification

- [x] `cargo test -p ferrocat-po --all-features --locked`
- [x] `cargo clippy -p ferrocat-po --all-targets --all-features --locked -- -D warnings`
- [x] `git diff --check`

## Benchmark note

- No local before/after timing report is included in this PR. The PR benchmark workflow should compare `rust-scheduled-v1` against `main`; keep #188 open after this PR because it only covers the first small sort/dedupe allocation chunk.

## Deferred from #188

- Consuming `update_catalog` normalization and merge-index ownership.
- PO export and artifact-provenance restructuring.
- Broader catalog API benchmark work.

### DIFF
--- a/crates/ferrocat-po/src/api/catalog.rs
+++ b/crates/ferrocat-po/src/api/catalog.rs
@@ -657,19 +657,18 @@ pub(super) fn sort_messages(messages: &mut [CanonicalMessage], order_by: OrderBy
                 .then_with(|| left.msgctxt.cmp(&right.msgctxt))
                 .then_with(|| left.obsolete.cmp(&right.obsolete))
         }),
-        OrderBy::Origin => messages.sort_by(|left, right| {
+        OrderBy::Origin => messages.sort_unstable_by(|left, right| {
             first_origin_sort_key(&left.origins)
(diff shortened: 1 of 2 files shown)


swernerx:
Benchmark artifact reviewed: `rust-scheduled-v1` passed with 8 comparable scenarios, 0 failures, and 0 noisy skips.

Median elapsed deltas: catalog PO parse -2.0%, catalog FCL parse -0.1%, ICU nested parse -12.2%, PO merge +0.5%, owned PO parse -2.7%, borrowed PO parse +0.7%, PO stringify -5.3%, catalog ICU-heavy update +1.8%. The small update-path slowdown is below the 20% regression gate.

Issue #224, opened 2026-07-03
Epic: post-2.0 consolidation — release, guardrails, API window, perf tail, comparative reach


swernerx:
Tracking issue for the post-2.0 consolidation arc: everything that came out of the v2.0.0 review round (#185-#195), the follow-ups it spawned, and the guardrails that keep the results from eroding. Ordered so each phase makes the next one cheaper - finish a phase before starting the next unless noted.

## Phase 1 - Ship 2.1.0 (now)

The cleanup work is on main; the release should go out before anything else lands on top of it.

- [x] #223 - write the 2.0 -> 2.1 migration guide on the docs site
- [x] Backfill the breaking-change bullets manually into the 2.1.0 changelog entries in the release PR (the commits intentionally carry no `BREAKING CHANGE:` footers, so release-please won't generate them; apply the #187 superset rule: the umbrella entry lists everything its re-exports inherit)
- [x] Link the migration guide from the changelog and the GitHub release notes
- [x] Merge the release PR (#184) and verify the published crates

**Why first:** every further change on main makes the release diff bigger, and the migration guide is only cheap while the renames are fresh.

## Phase 2 - Guardrails (before more surface/number changes)

Small, self-contained, and they make every later phase reviewable instead of trust-based.

- [x] #220 - public-API surface snapshot in CI (so the Phase 3 API changes show up as plain-text surface diffs in review)
- [x] #221 - published-numbers check against the checkpoint reports (so the Phase 4/5 benchmark refreshes can't leave stale README/homepage values), including rounding the fast-aging homepage proof stats

**Why second:** both are one-PR items with no design risk, and they change the character of Phases 3-5 from "be careful" to "CI has your back".

## Phase 3 - Close the API window (2.2)

The remaining reviewed breaking cleanups, done while downstream exposure is still effectively just Palamedes. After this phase the public surface should be boring.

- [x] #217 - fold `icu_options` / `syntax_policy` into the (now non-exhaustive) options structs; deprecate or remove the `_with_*` variants
- [x] #216 - decide the typed error/diagnostic-code contract; if the `DiagnosticCode` newtype direction is confirmed, implement it (serialize-transparent, JSON reports unchanged) and document the string-code stability contract in the `diagnostic_codes` module docs
- [x] Update the migration guide with anything user-visible from this phase

**Why third:** #217 becomes strictly harder after more adoption, and #216's outcome shapes how Palamedes consumes diagnostics - both want settling before the API is advertised as stable.

[...]

PR #225, opened 2026-07-03
Complete post-2.1 consolidation guardrails

### LINKED ISSUE #216: Decide typed API error and diagnostic-code contracts
## Summary

The broad #192 API cleanup tracker is now mostly resolved through PR #214 and PR #215. The remaining question is whether Ferrocat should add machine-readable typed error and diagnostic-code APIs instead of exposing only string payloads at the public boundary.

This issue tracks that decision separately so the original cleanup tracker can close.

## Current state

- `ApiError` is a small non-exhaustive enum with display/source-oriented variants such as `InvalidArguments(String)`, `Conflict(String)`, and `Unsupported(String)`.
- Diagnostic report types expose stable-looking machine codes as `String` fields, for example `Diagnostic.code`, `CatalogAuditDiagnostic.code`, and `CompiledCatalogDiagnostic.code`.
- The `diagnostic_codes` modules provide constants, but callers comparing report fields still compare plain strings.

## Design question

Should Ferrocat introduce typed public diagnostic/error identifiers, and if yes, what compatibility contract should they have?

Possible directions:

- keep the current string-based report shape and document the constants as the stable API
- add a lightweight `DiagnosticCode` newtype while preserving serialized JSON report strings
- add typed sub-kind enums for `ApiError` variants, for example conflict/invalid-argument categories, while keeping human-readable messages
- defer typed errors until the crate has more downstream consumers and clearer matching needs

## Constraints

- Do not break serialized report compatibility casually; diagnostics are likely to be consumed by CI and tooling.
- Avoid an ad hoc enum taxonomy that only mirrors current error text.
- Preserve useful `Display`, `Error::source`, and `ApiError::path()` behavior.
- Keep room for new diagnostic codes and error categories without forcing exhaustive downstream matches.

## Acceptance criteria

- Decide whether typed diagnostic/error identifiers are worth adding now.
- If yes, propose the concrete public shapes and serde behavior before implementation.
- If no, document the current string-code contract clearly enough that downstream tooling can rely on it deliberately.

Split from #192 after PR #214 and PR #215 resolved the other API cleanup items.

### LINKED ISSUE #217: API: consolidate _with_icu_options/_with_syntax_policy variants into the options structs
## Problem

The `_with_*` entry-point duplication flagged in #190 survived the cleanup round (PRs #208/#213 delivered the `#[non_exhaustive]` + builder convention, but did not consolidate the parallel functions):

- `audit_catalogs` vs. `audit_catalogs_with_icu_options` (`crates/ferrocat-po/src/api/audit.rs`)
- `compile_catalog_artifact` vs. `compile_catalog_artifact_with_icu_options` (`api/compile.rs`)
- `compile_catalog_artifact_selected` vs. `compile_catalog_artifact_selected_with_icu_options`
- the pseudolocalization pair with `_with_syntax_policy` (`api/pseudolocalization.rs`)

No `#[deprecated]` markers exist anywhere in the workspace, so all variants look equally canonical to a new user.

## Why now

The blocker that originally forced the parallel functions is gone: since PR #213 the main options structs are `#[non_exhaustive]` with `new()`/`with_*` builders, so they can grow an `icu_options` (resp. `syntax_policy`) field additively. `compile_catalog_artifact_report` with `CompileCatalogArtifactReportOptions` (which already embeds `icu_options` and a `selection` enum) remains the reference shape.

The project currently allows reviewed breaking cleanups in minor releases (see the Compatibility Snapshot in the README), and downstream exposure is effectively just Palamedes — the migration cost only grows from here.

## Proposed fix

1. Add `icu_options: CatalogAuditIcuOptions` to `CatalogAuditOptions` (default: current behavior) with a `with_icu_options(...)` builder; same for the compile and pseudolocalization options.
2. Route the `_with_*` variants through the new fields and mark them `#[deprecated]` with a pointer to the builder — or remove them outright in the same cleanup window, consistent with how the #192 renames were handled (hard cut, changelog-documented).
3. Update rustdoc examples and the docs-site API overview to show only the consolidated entry points.

## Outcome

One function per operation, options express the variability — the compile surface shrinks from four public functions to two (`compile_catalog_artifact`, `compile_catalog_artifact_report`), and future knobs no longer multiply the function count.

Follow-up to #190 (the remaining unimplemented item from that issue).

### LINKED ISSUE #218: Performance: remaining allocation cleanups in the catalog API layer (helpers set path, dedup index, FxHash)
## Summary

Leftover allocation cleanups from #188 that the perf PR series (#201–#211) did not pick up. Individually small; listed here so they don't get lost. Measured context: the main #188 work already brought `po-update/catalog-icu-heavy` from 57 to 70 MiB/s (+23%, see the closing comment on #188) — these are the remaining single-site items in the same layer.

### 1. `merge_unique_strings` / `merge_unique_origins`: set path still clones per element

`crates/ferrocat-po/src/api/helpers.rs:66-71` (and the same pattern in `merge_unique_origins`, `helpers.rs:106-114`): the ≥8-element path builds `target.iter().cloned().collect::<BTreeSet<_>>()` — cloning every existing string — and then clones each incoming value again via `seen.insert(value.clone())` before pushing it. A `BTreeSet<&str>`/`HashSet<&str>` over borrowed keys covers the membership check without any cloning; incoming values only need to move into the target vec.

### 2. `push_normalized_message` keeps a cloning dedup index

`crates/ferrocat-po/src/api/catalog.rs:284, 357-365`: the dedup index is still `BTreeMap<(String, Option<String>), usize>` populated with `msgid.clone()`/`msgctxt.clone()` per extracted message — the same pattern PR #202 removed from `merge_catalogs` (which now uses a borrowed-key `BTreeMap<(&str, Option<&str>), usize>`, `catalog.rs:409-417`). The same borrowed-index approach applies; the index is dropped before the messages are consumed.

### 3. SipHash everywhere; `rustc-hash` never landed

The new `msgid_hash_index` from PR #210 hashes with `std::hash::DefaultHasher` (SipHash, `api/types.rs:863-867`), and the merge/dedup indexes use default-hasher `HashMap`s/`BTreeMap`s. For short-string keys hashed once per entry per operation, `rustc_hash::FxHashMap`/`FxHasher` is a low-risk 2-3× win on the hashing itself. `rustc-hash` is a tiny, `no_std`-compatible, widely-vetted dependency (used by rustc, OXC — the project's declared MSRV alignment peer). Suggested scope: the hash index in `types.rs`, the merge index in `catalog.rs`, and the borrow-based index in `src/merge.rs:156-165`.

## Verification

Each item should show up (modestly) in `ferrocat-bench compare rust-scheduled-v1` on the `po-update` and `po-merge` scenarios; the PR benchmark workflow covers regression safety.

Follow-up to #188.

### LINKED ISSUE #220: CI: public-API surface snapshot so re-export drift shows up in PR diffs
## Problem

The #185/#186 class of bug — umbrella re-exports drifting from the sub-crates, types reachable in `ferrocat-po` but not in `ferrocat` — was only caught by a manual review. CI now runs `cargo-semver-checks` (the "Semver compatibility" jobs), but that tool answers a different question: *did the surface change incompatibly vs. the published version?* It does not catch **additive gaps** (a new `ferrocat-po` item that was never re-exported), dual-canonical-path drift, or a type that becomes unnameable through the umbrella.

Since the project now explicitly allows reviewed breaking cleanups in minor releases, every surface change should be *visible in the PR diff* rather than discovered later.

## Proposed fix

Add a public-API snapshot to the repo and CI:

1. Generate the surface with [`cargo public-api`](https://crates.io/crates/cargo-public-api) for `ferrocat`, `ferrocat-po`, and `ferrocat-icu` and check the sorted listings in as text files (e.g. `crates/<name>/tests/public-api/<name>.txt` or a top-level `api-snapshots/`).
2. A CI step (or `#[test]` via the `public-api` crate) regenerates and diffs them; any change fails until the snapshot is updated in the same PR.
3. Reviewers then see every added/removed/renamed public item as a plain-text diff line — re-export gaps become impossible to miss, and the changelog writer has a mechanical checklist.

Nightly-toolchain note: `cargo public-api` needs a nightly for rustdoc JSON; pin one in the workflow the same way docs.rs does, or run it in a dedicated CI job so MSRV jobs stay untouched.

## Outcome

Umbrella drift (the #186 category) becomes structurally detectable instead of review-dependent, and the "breaking cleanups in minors" policy gets the tooling that keeps it honest.

### LINKED ISSUE #221: CI: verify published performance numbers against the checked-in checkpoint reports
## Problem

The published performance numbers now live in three places: the README table, the docs landing page (`docs/app/routes/home.tsx`, `parseCompare`/`updateCompare` arrays), and the checked-in checkpoint reports under `benchmark/results/`. They were just re-synced by hand (README 192→245 MiB/s, homepage charts, new full-catalog-update row) — and hand-synced numbers drift. The README review that started this cleanup round found exactly that: stale claims nobody updated (#194).

Two flavors of the problem:

1. **Chart/table values vs. report values.** Nothing verifies that the `rate:` values in `home.tsx` or the README table cells match the checkpoint report they claim to come from.
2. **Fast-aging precision in prose.** The homepage proof block still hardcodes "60 conformance cases / 454 assertions", and the performance footnote carries "362 MiB/s" / "1.16 GiB/s" — the same class of numbers the README already rounded away in #194.

## Proposed fix

1. Add a small checked-in mapping (e.g. `benchmark/published-numbers.json`) that declares which scenario in which checkpoint report backs each published value, and a script (`scripts/check-published-numbers.mjs` or a Rust test in `ferrocat-bench`) that fails CI when a published value deviates from its source by more than a rounding tolerance.
2. Wire it into the docs-build or lint CI job — cheap, no benchmark run needed, it only reads the JSON reports.
3. While in there: round the homepage proof stats to "60+" conformance cases (drop the assertion count or feed it from the conformance report), and source the owned-parse/serialize footnote numbers from the report or round them.

## Outcome

"Every published number traces to a checked-in report" stops being a discipline and becomes a check — the next perf gain (or regression) can't leave stale marketing numbers behind.

### LINKED ISSUE #222: Benchmark: add Babel (and evaluate Lingui) to the catalog-update cross-tool comparison
## Problem

The new `po-update` cross-tool groups (PR #219) compare the full `update_catalog` path against msgmerge, pofile, pofile-ts, and polib. Two comparison targets are missing that would make the story stronger and fairer:

1. **Babel's real `Catalog.update`** — the Python adapter already implements it (used by the `catalog-update-v1` profile with `no_fuzzy_matching=True`), but it is not wired into the ecosystem `po-update` groups. Babel is the closest thing Python has to a *catalog-maintenance* API, so it is the most honest per-language counterpart to `update_catalog` — closer than polib's plain merge.
2. **A JS extraction-pipeline tool** (e.g. Lingui's catalog merge step) — the tool a JS team would actually run for "update my catalogs with newly extracted strings". This is the comparison that matters for the Palamedes story. Needs evaluation first: Lingui's merge is entangled with its extractor, so the benchmark task must be cut carefully (feed pre-extracted messages, measure only the catalog update+write) to stay file-to-file honest.

## Proposed fix

1. Add `babel` scenarios to the `po-update/gettext-ui-de-{1000,10000}` groups in `gettext-workflows-ecosystem-v1` (adapter support exists; verify the semantic digest passes the same compatibility gate as msgmerge).
2. Evaluate a Lingui-based update implementation for the Node adapter; if the task can be cut cleanly, add it to the same groups — if not, document why in the benchmarking page (same transparency as the `gettext-parser` exclusion note).
3. Refresh the publication checkpoint and, if the numbers change the table, update README/homepage via the published-numbers process.

## Outcome

The update comparison covers each ecosystem's *best* maintenance path, not just its parser — which is exactly the claim the landing page makes.

swernerx:
## Summary

Completes the post-2.1 consolidation batch tracked in #224:

- Adds CI guardrails for the public Rust API surface and README/docs benchmark numbers.
- Hardens the publish workflow crate-existence probe after the trusted publishing rerun issue.
- Consolidates ICU option entry points into option structs and keeps diagnostics JSON-compatible through string-backed diagnostic code newtypes.
- Documents the 2.2 API cleanup migration path in the upgrade guide.
- Reduces catalog merge/update allocation work with borrowed indices and hash maps.
- Adds Babel update-workflow benchmark scenarios, documents why Lingui remains out of scope for isolated catalog-update comparison, and refreshes published benchmark evidence.

## Issue Status

- Closes #216
- Closes #217
- Closes #218
- Closes #220
- Closes #221
- Closes #222
- Updates #224

## Benchmark Evidence

Regression profile before/after for the merge-allocation work:

- `po-merge/mixed-10000/ferrocat`: 5.955 -> 5.664 ms/iter (-4.88%), non-noisy.
- `po-update-file/gettext-ui-de-10000/ferrocat`: 19.817 -> 17.259 ms/iter (-12.91%), non-noisy.
- `po-update/catalog-icu-heavy/ferrocat`: 3.012 -> 2.254 ms/iter (-25.17%), baseline noisy, current non-noisy.
- Regression check: 8 passed, 0 failed, 1 skipped-noisy.

Updated ecosystem checkpoint: `benchmark/results/gettext-workflows-ecosystem-v1-255727f.json`.

## Validation

- [x] `bash scripts/check-public-api-snapshots.sh`
- [x] `node scripts/check-published-numbers.mjs`
- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps --locked`
- [x] `cargo +1.93.0 check --workspace --all-targets --all-features --locked`
- [x] `cargo llvm-cov --workspace --all-features --locked --lcov --output-path target/lcov.info --ignore-filename-regex 'crates/ferrocat-(bench|conformance)/'`
- [x] `cargo llvm-cov report --json --summary-only --output-path target/coverage-summary.json --ignore-filename-regex 'crates/ferrocat-(bench|conformance)/'`
- [x] `node scripts/coverage-gate.mjs target/coverage-summary.json ferrocat-po=95 ferrocat-icu=95`
- [x] `(cd docs && pnpm install --frozen-lockfile)`
- [x] `(cd docs && pnpm build)`
- [x] `cargo run --release -p ferrocat-bench -- compare gettext-workflows-ecosystem-v1 --out benchmark/results/gettext-workflows-ecosystem-v1-255727f.json`


swernerx (commented):
Review pass over the whole batch. The option-struct consolidation and the merge-allocation rewrite hold up well — the select-then-move dedup trick reads harder than the old code, but the doc comments carry it. The new guard rails (API snapshots, published-numbers check) are a real upgrade for the repo.

One actual bug (a mangled sentence on the landing page), two robustness items in CI/publish, and a couple of API-shape thoughts — all inline below. One extra thought that didn't fit a line: the select-then-move dedup pattern now exists three times (twice in `helpers.rs`, once in `catalog.rs`). Fine for now, but if a fourth copy shows up, extract a shared helper keyed on a closure.

docs/app/routes/home.tsx
```diff
@@ -453,9 +453,10 @@ export default function HomePage() {
           the high-level catalog update on the same files: on top of the plain
           update it analyzes ICU message structure, tracks placeholders, and
           produces deterministic output. None of the compared tools has an
-          equivalent layer&mdash;their update <em>is</em> the plain bar&mdash;
-          and a compatibility probe asserts the output stays semantically
-          identical to msgmerge&rsquo;s on this corpus. The GNU msgmerge bar is not
+          broader benchmark now includes Babel&rsquo;s real Python
+          <code>Catalog.update</code> path, and a compatibility probe asserts
```
swernerx, line 458: This sentence got mangled in the edit — it now renders as "None of the compared tools has an broader benchmark now includes Babel's real Python Catalog.update path". The "equivalent layer" clause lost its ending when the Babel mention was spliced in. Something like:

```
... None of the compared tools has an
equivalent layer&mdash;their update <em>is</em> the plain bar. The
broader benchmark now includes Babel&rsquo;s real Python
<code>Catalog.update</code> path, and a compatibility probe asserts ...
```

This is on the landing page, so it should be fixed before merge. Also a good reminder that the published-numbers check only asserts substrings exist — it can't catch this class of prose breakage.

.github/workflows/publish.yml
```diff
@@ -172,22 +172,8 @@ jobs:
...
-                ;;
-              404)
-                return 1
-                ;;
-              *)
-                echo "Unexpected crates.io response ${status} while checking ${crate} ${version}."
-                return 1
-                ;;
-            esac
+            cargo search "$crate" --limit 20 | grep -F "${crate} = \"${version}\"" > /dev/null
```
swernerx: Two edge cases where `cargo search` gives the wrong answer: it only surfaces the *latest* version of a crate, and it hides yanked crates. So re-running a publish for an older tag after a newer release went out (or after a yank) reports "not published", and the workflow attempts a republish that crates.io rejects — which is exactly the rerun scenario this probe exists for. The sparse index is authoritative and lists every version, including yanked ones:

```bash
curl -fsSL "https://index.crates.io/${crate:0:2}/${crate:2:2}/${crate}" \
  | grep -F "\"vers\":\"${version}\"" > /dev/null
```

All four crate names are 4+ chars, so the two/two prefix path always applies. No search-index coupling, no crates.io API rate-limit or user-agent surprises.

.github/workflows/ci.yml
```diff
@@ -145,6 +148,26 @@ jobs:
...
+  public-api:
+    name: Public API snapshot
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v7
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@nightly
```
swernerx: #220 called out pinning the nightly, and it ended up mattering within this very PR — the `core::io::error::Error` sed two commits later is exactly the kind of nightly-dependent rustdoc output this job is exposed to. Left unpinned, any nightly bump can start failing snapshot checks on PRs that touched nothing. I'd pin it (`toolchain: nightly-2026-06-xx` via the action's `with:`) right next to the pinned cargo-public-api 0.52.0, and bump both deliberately.

crates/ferrocat/src/lib.rs
```diff
@@ -196,21 +194,19 @@ pub use ferrocat_po::{
...
-    compile_catalog_artifact, compile_catalog_artifact_report, compile_catalog_artifact_selected,
-    compile_catalog_artifact_selected_with_icu_options, compile_catalog_artifact_with_icu_options,
-    compiled_key, machine_translation_hash, measure_catalog_coverage, parse_catalog,
-    pseudolocalize_compiled_catalog_artifact,
-    pseudolocalize_compiled_catalog_artifact_with_syntax_policy, review_catalogs, update_catalog,
-    update_catalog_file,
+    CompiledCatalogPseudolocalizationOptions, CompiledCatalogResolution,
+    CompiledCatalogResolutionKind, CompiledCatalogTranslationKind, CompiledCatalogUnavailableId,
+    CompiledKeyStrategy, CompiledMessage, CompiledTranslation, DescribeCompiledIdsReport,
+    Diagnostic, DiagnosticSeverity, EffectiveTranslation, EffectiveTranslationRef,
```
swernerx: `Diagnostic`, `CatalogAuditDiagnostic`, and `CompiledCatalogDiagnostic` are all re-exported at the top level, and their `code` field is now a `DiagnosticCode` — but the type itself is only nameable as `ferrocat::catalog::DiagnosticCode`. Anyone working from the top-level exports can't spell the field's type without dropping into a module. That's the #186 drift pattern the new snapshot job is built to surface (and it does — the gap is plainly visible in `api-snapshots/ferrocat.txt`), so let's close it here: add `DiagnosticCode` to this list.

crates/ferrocat-po/src/diagnostic_codes.rs
```diff
@@ -6,6 +6,80 @@
...
+
+/// Stable machine-readable diagnostic code.
+///
+/// The wire format remains the plain code string. This wrapper makes diagnostic
+/// code fields distinct from arbitrary human-readable strings while preserving
+/// string comparisons through [`Self::as_str`], [`AsRef<str>`], and `PartialEq<&str>`.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+pub struct DiagnosticCode(String);
```
swernerx: This is a byte-for-byte copy of the `DiagnosticCode` in `ferrocat-icu`, and the umbrella crate now exposes both as distinct types with the same name (`catalog::DiagnosticCode` vs `icu::DiagnosticCode`). The friction is already visible in this PR: `audit.rs` and `compile.rs` bridge between them with `.as_str()` / `.as_str().into()` round-trips. Since `ferrocat-po` already depends on `ferrocat-icu`, re-exporting the icu type here would collapse them into one type and drop those conversions — the serde gate carries over by forwarding the feature. Not blocking (the JSON contract is identical either way), but I'd rather not let two same-named public types calcify into the 2.2 surface, because unifying them later is another breaking change.

crates/ferrocat-po/src/api/compile_types.rs
```diff
@@ -226,6 +239,19 @@ pub struct CompileCatalogArtifactIcuOptions {
     pub formatter_support: Option<IcuFormatterSupportPolicy>,
 }
 
+impl PartialEq for CompileCatalogArtifactIcuOptions {
```
swernerx, line 242: Worth a doc note on this impl: `fn_addr_eq` is best-effort by language rules — the same function can compare unequal across codegen units (false negatives), and distinct-but-identical functions can be merged and compare equal. Totally fine for the purpose (the `PartialEq` derive on `CompileCatalogArtifactOptions` needs *something* here), but the caveat should live next to the code so nobody later builds logic on exact callback identity.

docs/app/routes/home.tsx
```diff
@@ -225,19 +224,20 @@ const parseCompare: Bar[] = [
...
-    rate: 88,
+    rate: 108,
     self: true,
   },
-  { name: "pofile-ts", lang: "Node", rate: 39 },
+  { name: "pofile-ts", lang: "Node", rate: 43 },
   { name: "gettext/gettext", lang: "PHP", rate: 16 },
-  { name: "polib", lang: "Python", rate: 7 },
-  { name: "msgmerge", lang: "GNU gettext", rate: 4.5 },
+  { name: "Babel", lang: "Python", rate: 8 },
```
swernerx: Small fairness footnote, take it or leave it: this bar is Babel's `po-update` number, while the pofile-ts/polib/msgmerge bars in the same chart come from the `po-merge` groups. The externals barely differ between the two (pofile-ts 43 vs 42), so the picture stays honest — but the second Ferrocat bar gets an explicit "full catalog update" label while Babel's update-path number sits here unlabeled. A short `lang` note like "Python, Catalog.update" would make it airtight.

swernerx:
Review feedback addressed in `ed02f1b`.

Covered items:
- fixed the mangled landing-page sentence and labeled Babel as `Python, Catalog.update` in the update chart
- replaced the publish rerun probe with the crates.io sparse index so older/yanked versions are detected
- pinned the public API snapshot job to `nightly-2026-06-30` and wired the snapshot script through `PUBLIC_API_TOOLCHAIN`
- re-exported `DiagnosticCode` at the umbrella top level and unified `ferrocat-po` catalog diagnostics with the `ferrocat-icu` `DiagnosticCode` type
- documented the best-effort nature of callback address equality near the `ptr::fn_addr_eq` use

Local validation for the follow-up commit:
- `cargo fmt --all --check`
- `node scripts/check-published-numbers.mjs`
- `cargo test -p ferrocat-icu -p ferrocat-po -p ferrocat --all-features --locked`
- `cargo check -p ferrocat-po --no-default-features --features serde --locked`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p ferrocat-icu -p ferrocat-po -p ferrocat --all-features --no-deps --locked`
- `(cd docs && pnpm build)`

CI is running on the pushed branch now.

swernerx:
Follow-up status: CI is green after `ed02f1b`, Codecov patch/project are green, and all addressed review threads are resolved. PR merge state is `CLEAN`.

crates/ferrocat-po/src/diagnostic_codes.rs
swernerx: Resolved in ed02f1b4 via the conditional re-export — with `catalog` on, both crates now share one type, and the no-catalog fallback copy is compile-checked by the documented-feature-profiles CI step, so both paths stay honest.

One residual caveat worth a line of comment in the code: the type *identity* of `ferrocat_po::DiagnosticCode` now depends on the `catalog` feature. Strictly speaking that makes the feature non-additive — a downstream that builds without `catalog` and implements a trait for both `ferrocat_po::DiagnosticCode` and `ferrocat_icu::DiagnosticCode` would hit a coherence error the moment anything in its graph enables `catalog`. With the current downstream surface that's theoretical, but the cfg switch is non-obvious enough that a short note explaining why the fallback copy exists (and that identity merges under `catalog`) would save the next reader the archaeology.

swernerx: Comment added in cdf23ad — the cfg switch now documents the shared-vs-fallback split and the feature-dependent type identity.

PR #242, opened 2026-07-10
perf: tighten audits, benchmarks, and project safeguards

### LINKED ISSUE #236: perf(audit): reuse parsed ICU messages across syntax and compatibility checks
## What I found

The default `CatalogAuditChecks` enables both ICU syntax validation and ICU compatibility checks. Those two passes parse the same messages independently.

Current flow in `crates/ferrocat-po/src/api/audit.rs`:

- `audit_catalog_entries` parses every active source message and every active target translation for syntax.
- `audit_icu_compatibility` then parses each source `msgid` again for every target locale.
- The same target translation is parsed again before compatibility comparison.

That makes the default path do roughly `source_messages + 2 × target_messages + source_messages × target_locales` ICU parses. The extra source pass is especially noticeable as locale count grows.

## Measured impact

I ran a release-mode probe with 10,000 messages and five target locales, using seven runs and the median:

- syntax + compatibility: 52.2 ms
- syntax only: 16.2 ms
- compatibility only: 40.0 ms

This isn't a publication benchmark, but it confirms the duplicated parsing is a meaningful part of the audit cost. The control flow also makes the repeated work deterministic, not a profiler guess.

## Suggested direction

Parse each active source ICU message once, then reuse that result across all target locales. Within a target locale, reuse the translation parse between syntax validation and compatibility comparison.

A small internal cache keyed by `CatalogMessageKey`, or a combined audit pass that carries parsed values forward, should work. Keep parse errors and `IcuSyntaxPolicy` behavior exactly as they are today.

## Done when

- Default audits don't reparse the same source message per target locale.
- Target translations aren't parsed once for syntax and again for compatibility.
- Diagnostics and ordering stay unchanged.
- A multi-locale audit scenario covers the change in the benchmark harness.

### LINKED ISSUE #237: bench: cover catalog QA and runtime workflows in the PR regression profile
## Gap

The PR-visible `rust-scheduled-v1` profile currently covers only these operations:

- `parse`
- `stringify`
- `merge`
- `update-catalog`
- `parse-catalog`
- `parse-icu`

That leaves several public, catalog-sized workflows without a same-run regression guard:

- `combine_catalogs`
- `audit_catalogs`
- `measure_catalog_coverage`
- `review_catalogs`
- `compile_catalog_artifact` and report/selected variants

There is a standalone `combine-catalogs` microbenchmark, but the compare runner doesn't support it, so it isn't part of PR baseline/current comparisons. The other workflows aren't represented in `ferrocat-bench` at all.

This also makes the README claim at line 73 too broad: it says combine, audit, and runtime paths are covered by PR-visible benchmark regression checks.

## Why it matters

These APIs walk full catalogs and often multiply work by locale count. They can regress through extra parsing, cloning, indexing, or report construction without touching the parser benchmarks that CI watches today.

## Suggested scope

- Add internal compare-runner operations for combine, audit, coverage, review, and artifact compilation.
- Use deterministic multi-locale fixtures, including a 10k-message case.
- Validate each output with a semantic digest before measuring it.
- Put a representative subset in `rust-scheduled-v1`; keep broader variants in a slower profile if needed.
- Tighten the README wording if any public path intentionally stays outside the PR guard.

## Done when

A PR that slows one of these catalog-sized workflows can trip the same baseline/current regression check used by parser and update scenarios.

### LINKED ISSUE #238: refactor(bench): split compare.rs into focused benchmark modules
## Problem

`crates/ferrocat-bench/src/compare.rs` has grown to 4,232 lines and about 110 top-level items. It currently owns all of these concerns in one module:

- CLI option parsing
- profile loading and validation
- scenario preparation and scheduling
- calibration and execution
- internal and external adapters
- semantic normalization and digests
- report schemas and statistics
- regression comparison
- host and toolchain detection
- roughly 900 lines of tests

The file now carries three `clippy::too_many_lines` expectations. The comments explain why each routine is centralized, but the module itself no longer has a single responsibility.

## Why it matters

Adding a new benchmark operation means touching preparation, dispatch, validation, artifacts, reporting, and tests in one very large file. That makes review harder and increases the chance that an operation gets wired into only part of the runner.

This is already relevant for adding catalog QA and runtime workflows to the PR profile.

## Suggested split

Keep the scheduling algorithm cohesive, but move the surrounding responsibilities into focused private modules, for example:

- `profile`
- `runner`
- `operations`
- `semantic_digest`
- `report`
- `regression`
- `environment`
- `adapters`

The exact names don't matter. The useful boundary is that adding an operation shouldn't require navigating a 4k-line file.

## Done when

- `compare.rs` becomes a small command/orchestration entry point.
- Existing report JSON stays compatible.
- Round-robin scheduling, calibration, semantic validation, and regression tests keep the same behavior.
- No benchmark result changes unless separately measured and documented.

### LINKED ISSUE #239: ci: make semver compatibility checks blocking again
## Problem

The SemVer matrix in `.github/workflows/ci.yml` still runs `cargo-semver-checks` with `continue-on-error: true`.

The nearby comment says this is for the planned 2.1 cleanup window. That window has shipped, and the repository is now on 2.2.0, but the exception is still active. A new incompatible public API change can therefore fail all three SemVer jobs without blocking the PR.

The public API snapshot job is useful, but it doesn't replace semantic compatibility checking: a snapshot can be updated in the same PR even when the change is accidentally breaking.

## Fix

- Remove `continue-on-error: true`.
- Remove or update the stale 2.1 cleanup comment.
- If a future release intentionally breaks API, handle that as an explicit major-release decision instead of leaving the guard globally non-blocking.

## Done when

A SemVer failure for `ferrocat-icu`, `ferrocat-po`, or `ferrocat` makes CI fail again.

### LINKED ISSUE #240: docs: update the release and upgrade guidance for the shipped 2.2 line
## What is stale

The release guidance still reads like the 2.1/2.2 cleanup work hasn't shipped:

- `docs/app/routes/guide/project-status.mdx` calls 2.1.0 a planned cleanup window in both the intro and SemVer policy.
- `docs/app/routes/guide/upgrading.mdx` does the same.
- The upgrade page still has an `Upcoming 2.2 API cleanup` section even though the option consolidation and diagnostic-code work are already in the released history.
- The lockstep dependency examples still use `ferrocat = "1"`, `ferrocat-po = "1"`, and `ferrocat-icu = "1"` while the current stable line is 2.2.

## Why it matters

This page is the source of truth for compatibility expectations. Right now it mixes historical exceptions, future tense, and an older dependency line, so a new user can't tell which migration steps are already required.

## Update

- Move the 2.1 cleanup language to past tense.
- Reconcile the API cleanup section with the release/changelog where it actually shipped, then rename it from `Upcoming`.
- Update the dependency examples to the current major line, or make them explicitly version-neutral.
- Keep the temporary 3.0/yank history clear if it affects what users may have resolved.
- Align the Project Status SemVer wording with the now-closed cleanup window.

## Done when

The page clearly separates current policy from historical migration notes and matches the published 2.2 API.

### LINKED ISSUE #241: docs: keep conformance and coverage snapshots in sync with CI
## Drift found

The quality pages present exact repository metrics, but those numbers are already behind the current code.

`docs/app/routes/quality/conformance.mdx` says:

- 60 cases
- 454 assertions
- 55 expected passes

The current `ferrocat-bench conformance-report` output is:

- 61 cases
- 457 assertions
- 56 expected passes
- 5 expected rejects
- 0 known gaps

The coverage page is also behind the current CI command. A fresh `cargo llvm-cov` summary reports:

- `ferrocat-cli`: 87.13% (docs: 87.32%)
- `ferrocat-po`: 97.40% (docs: 96.82%)
- `ferrocat-icu`: 96.85% (docs: 96.98%)

The thresholds are still correct; the exact snapshots are the drifting part.

## Suggested fix

Either generate these sections from machine-readable reports, or add a cheap CI check that compares the documented values with:

- `cargo run -p ferrocat-bench -- conformance-report`
- `target/coverage-summary.json` plus `scripts/coverage-gate.mjs`

If exact coverage percentages are intentionally just dated snapshots, label them as historical and avoid calling the section `Current Gate`. The conformance counts should still have a source-of-truth check because they are cheap to compute.

This is the same class of drift that #221 fixed for published benchmark numbers, but these quality pages aren't covered by that script.

## Done when

A new conformance case or a meaningful coverage refresh can't leave the quality docs silently stale.

swernerx:
## Summary

This PR closes the performance, structure, CI, and documentation maintenance batch:

- [x] #236 — reuse parsed ICU messages during catalog audits (`6be91e3`)
- [x] #237 — add catalog workflow regression scenarios to the benchmark runner (`f2e65ff`)
- [x] #238 — split the benchmark comparison harness into focused modules (`c74af41`)
- [x] #239 — make semver compatibility checks blocking again (`7f3d25e`)
- [x] #240 — align the upgrade guide and project status with the 2.2 release line (`6ee1640`)
- [x] #241 — guard published conformance and coverage claims against drift (`9026468`)
- [x] Cover syntax-only and compatibility-only audit cache fallback paths (`2259c77`)
- [x] Restore Classic Gettext mode for mixed benchmark update fixtures (`99f4b98`)
- [x] Refresh the external benchmark publication checkpoint (`7d470df`)

The implementation plans are in `docs/specs/2026-07-10-issues-236-241-maintenance.md` and `docs/superpowers/specs/2026-07-11-benchmark-checkpoint-design.md`.

## Performance

The audit path now shares parsed source ASTs across target locales and reuses parsed target ASTs between syntax and compatibility checks. The cache stays local to one audit invocation.

A standalone release-mode probe with 10,000 complex ICU messages and five target locales measured:

- before: 131.864 ms median
- after: 116.493 ms median
- improvement: about 11.7%
- diagnostics: unchanged at zero

The expanded `rust-scheduled-v1` profile covers 14 scenarios, including combine, audit, coverage, review, and compiled-artifact workflows.

## Benchmark checkpoint

Fresh release-mode reports ran on Apple M1 Ultra with the documented toolchain and fixture matrix:

- parse: Ferrocat borrowed parsing 695 MiB/s; pofile-ts 148 MiB/s
- plain catalog update: Ferrocat 255 MiB/s; pofile-ts 42 MiB/s
- full catalog update: Ferrocat 109 MiB/s; pofile-ts 42 MiB/s, Babel 7.9 MiB/s, GNU msgmerge 4.6 MiB/s adjusted
- the full update remains about 2.6x ahead of the fastest external workflow and roughly 24x ahead of GNU msgmerge
- FCL catalog parsing is about 25% faster than equivalent PO in the current storage-format profile

The PR adds the official, workflow, compatibility, and catalog-update reports. Rust-only scheduled, storage-format, and serious-profile reports remain CI/local artifacts. The serious profile now completes all 37 scenarios; noisy lanes are retained in the report but excluded from public headline numbers.

## Structure and compatibility

The benchmark comparison entry point is down from 4,232 lines to 485 lines, with adapters, environment handling, operations, profiles, regression checks, reporting, semantic digests, and tests moved into focused modules.

No public API or serialized format changed. The semver jobs for `ferrocat`, `ferrocat-icu`, and `ferrocat-po` all pass as blocking checks.

## Verification

- [x] `cargo test -p ferrocat-bench --all-features --locked`
- [x] Docs install and production build
- [x] Published-number consistency check
- [x] Release-mode runs: official, workflows, compatibility, catalog update, storage formats, scheduled Rust, and serious
- [ ] GitHub Actions and Codecov for `7d470df`

Closes #236
Closes #237
Closes #238
Closes #239
Closes #240
Closes #241


crates/ferrocat-bench/src/compare.rs
```diff
@@ -466,3767 +500,4 @@ fn parse_positive_f64(label: &str, value: &str) -> Result<f64, String> {
...
-            compare_regression_reports(&baseline, &current, 20.0).expect("regression report");
-
-        assert!(
-            report.has_failures(),
-            "per-iteration slowdown must fail: {}",
-            report.render()
-        );
-    }
-}
+include!("compare/tests.rs");
```
swernerx: This `include!` keeps tests.rs invisible to rustfmt: fmt only walks real `mod` declarations, so `cargo fmt --all --check` passes even with deliberately mangled formatting in that file (tried it locally to be sure). Clippy still compiles the included code, so lints are covered; only formatting is unguarded. `#[cfg(test)] mod tests;` here, plus dropping the inner `mod tests { ... }` wrapper in tests.rs, gives you the same scoping with standard tooling. Worth doing while the split is fresh.

crates/ferrocat-po/src/api/audit.rs
```diff
@@ -333,31 +341,52 @@ pub fn audit_catalogs(
...
+                .icu_compatibility
+                .then_some(&mut source_icu_cache),
             &mut report,
         );
     }
     if options.checks.semantic_metadata {
         audit_metadata(options.metadata, &source_keys, &mut report);
     }
+    if options.checks.icu_compatibility {
+        cache_source_icu_messages(
```
swernerx, line 357: Small trade-off that deserves a one-line comment in the code: with `icu_syntax` disabled, this now parses every active source msgid up front, including keys no target ever translates. The old path parsed a source lazily, only once some target actually had a translation for it. Dense catalogs come out ahead, sparse compatibility-only audits pay a bit more. The default config is unaffected since the syntax pass already fills the cache. Fine as-is; making it lazy would need mutable access to the source cache while it's shared across the target loop, so the eager fill is the simpler shape. I'd just note that choice here so nobody optimizes it back later.

swernerx (commented):
Went through all six issue slices. This holds up well: I diffed the module split against the old compare.rs and the only deltas are visibility bumps to `pub(super)`, the audit cache falls back to a fresh parse on any miss so diagnostics can't drift, and the "shipped in 2.1.1" claim in the upgrade guide matches where the ICU option consolidation actually landed (checked the tags). The conformance-docs guard test is a nice touch; that page had no reason to keep drifting.

Two notes on the lines below, neither blocks. Would approve, but GitHub does not let this account approve its own PR.

PR #245, opened 2026-07-11
ci(release): consolidate GitHub product releases

swernerx:
## What changed

Ferrocat now keeps `ferrocat-vX.Y.Z` as the one visible GitHub product release. Release Please still creates the component tags it needs for bookkeeping, but a final publish job folds component release notes into the product release and removes the five redundant component release entries after the crates and CLI archive are published.

The musl CLI archive now lands on the Ferrocat product release instead of the `ferrocat-cli` release. The release verification docs and both READMEs now point to the product release too.

## Why

All workspace crates ship at the same linked version, so six GitHub releases per version made the Releases page noisy without adding useful release boundaries.

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/publish.yml"); puts "workflow YAML parses"'`\n- extracted cleanup script checked with `bash -n`\n- `pnpm build` in `docs/`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,9 +25,8 @@ jobs:
       pull-requests: write
     outputs:
       releases_created: ${{ steps.release.outputs.releases_created }}
-      ferrocat_cli_release_created: ${{ steps.release.outputs['crates/ferrocat-cli--release_created'] }}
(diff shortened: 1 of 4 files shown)


swernerx:
Closing this in favor of the single product-release spike. The cleanup worked, but keeping Release Please component-based and deleting releases afterward isn't the model we want to standardize.

PR #252, opened 2026-07-12
fix(release): synchronize product lockfile

swernerx:
## Summary

Fixes the product release PR path after the move to one root component.

- syncs `Cargo.lock` on the Release Please PR branch only when that PR is created or updated
- keeps registry dependency versions pinned; only the six local workspace package versions move
- reads the docs version from the root manifest key (`.`)

## Proof

A publishing-free Action run created one `2.3.0` product release PR, then added a bot commit with exactly the six expected `Cargo.lock` updates. No package was published.

## Validation

- `cargo fmt --all --check`
- `cargo check --workspace --all-targets --all-features --locked`
- `pnpm install --frozen-lockfile && pnpm build` in `docs/`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -42,6 +42,54 @@ jobs:
           manifest-file: .release-please-manifest.json
           token: ${{ secrets.RELEASE_PLEASE_TOKEN || secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


swernerx:
Superseded by #255. The root-package setup lets Release Please's native Rust strategy update Cargo.lock and the full workspace directly, so the follow-up workflow commit is no longer needed.

PR #268, opened 2026-07-29
feat(po): use CLDR root catalog ordering by default

### LINKED ISSUE #259: Let callers express a catalog order beyond `Msgid` and `Origin`
## Problem

`RenderOptions::order_by` offers `OrderBy::Msgid` and `OrderBy::Origin`. Both
sort by `String::cmp`, which is code-point order. A consumer that needs any
other order has no way to say so, and the only workaround is to re-parse and
re-serialize whatever Ferrocat just rendered.

That is what Palamedes does today, and it costs more than it should.

## Why Palamedes needs another order

Lingui orders catalogs with `new Intl.Collator("en-US")`, which resolves to the
unmodified CLDR root collation — English carries no collation tailoring of its
own. Palamedes wrote code-point order, so migrating a Lingui catalog re-sorted
it wholesale on the first extraction: roughly 678,000 changed lines across 20
files on the case that prompted this
(sebastian-software/palamedes#479).

The two orders differ far more than "a few accented words". Over all 80 example
catalogs in the Palamedes repo plus a corpus of realistic UI strings, ~27 % of
pairs are inverted between them, and the failures are systematic rather than
exotic: `{` versus `<` decides the relative order of a large share of extracted
messages — ICU MessageFormat placeholders against `<0>…</0>` markup — and
code-point order gets that backwards.

## What the workaround costs

Palamedes now renders through Ferrocat and then re-parses, re-orders and
re-serializes the result. Profiled on a catalog file of 4950 entries:

| phase | time |
| --- | --- |
| `parse_po` | ~1.2 ms |
| metadata restore | ~0.0 ms |
| sort | ~1.5 ms |
| `stringify_po` | ~0.5 ms |

The parse and stringify are pure overhead — roughly 1.7 ms of the 3.9 ms pass,
and Ferrocat already sorts internally before rendering. Doing it once inside
`update_catalog` would remove that entirely.

It also leaks in two places a post-pass cannot reach cleanly:

- **`combine_catalog_files` renders in code-point order.** Palamedes uses
  `pmds catalog merge` as a Git merge driver, so every resolved conflict came
  back fully re-sorted and the next extraction sorted it straight back. Patched
  in sebastian-software/palamedes#482 by post-processing the written file —
  which would be unnecessary if the order were a render option.

- **PO and FCL disagree.** The same six extracted messages:

  ```
  PO   Álgebra, apple, Apple, Uber, über, Zebra
  FCL  Apple, Uber, Zebra, apple, Álgebra, über
  ```

  `Apple` and `apple` sit at opposite ends of the FCL file. Palamedes only
  post-processes PO, and adding a second post-pass for FCL would repeat the
  mistake rather than fix it. A render option would cover both formats and the
  combine path at once.

## Proposal

Add a variant that carries the collation, rather than a caller-supplied
comparator:

```rust
pub enum OrderBy {
    Msgid,
    Origin,
    /// Sort by msgid and then msgctxt using the CLDR root collation.
    Collated,
}
```

A `Custom(Box<dyn Fn(&CanonicalMessage, &CanonicalMessage) -> Ordering>)` would
be less invasive for Ferrocat, but it pushes ordering semantics back out to
every caller — which is the situation this is meant to end. Catalog ordering is
catalog semantics.

## What could move over

Palamedes already carries a working implementation that could transfer whole:

- a generated table of primary weights plus canonical decompositions, derived
  from `Intl.Collator` itself by a script, with a `--check` mode that keeps the
  checked-in file current;
- a sort-key builder with an ASCII fast path and a packed 16-byte prefix so
  full keys are only built for entries that share one;
- a golden test against real Lingui 6 output — a fixture produced by `pofile`
  ordered with `Intl.Collator("en-US")`, matched byte for byte;
- end-to-end verification over ~5000-message catalogs, zero entries at a
  different position than `Intl.Collator("en-US")` puts them.

Two documented limits come with it, both outside what source messages hold in
practice: ligatures and digraphs (`ﬁ`, `Ǆ`) expand to several collation
elements in the real algorithm and a flat table cannot express that; characters
outside the covered repertoire sort after it by code point, so the placement of
non-Latin scripts after Latin holds but the order within them does not. Since
this only decides the order entries appear in, a miss costs a line in a diff
rather than a wrong translation — which is what makes a partial collation the
right trade here.

The size cost is the reason for a table rather than ICU4X: linking
`icu_collator` added 1,301,440 B to the release binary, against ~65,512 B for
the table, with identical output over the corpora above. That would land in
every Ferrocat consumer, so it is worth deciding deliberately — but at ~20 KB of
data rather than 1.3 MB.

## References

- sebastian-software/palamedes#479 — the migration case
- sebastian-software/palamedes#481 — collation order and the table
- sebastian-software/palamedes#482 — the merge-driver workaround

swernerx:
## Summary

- make the existing default `OrderBy::Msgid` use the CLDR root order from `Intl.Collator("en-US")`, with `OrderBy::Origin` using the same message/context tie-break
- apply the same identity order across PO update/combine output; make every new FCL file declare and use `order=collated` while retaining read compatibility for legacy tagless bytewise FCL
- derive compact primary and secondary weights from Intl, preserve accent positions and canonical precomposed/decomposed equivalence, and verify the generated encoding across 90,951 accent pairs
- add ASCII and 32-byte packed-prefix fast paths, borrowed raw tie-breakers, a byte-for-byte Lingui golden, FCL/order regressions, CI provenance checks, and ADR 0026
- keep the implementation dependency-free and document its deliberate ligature/digraph and out-of-repertoire limits

## Verification

- [x] `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps --locked`
- [x] `cargo +1.93.0 check --workspace --all-targets --all-features --locked`
- [x] `PUBLIC_API_TOOLCHAIN=nightly bash scripts/check-public-api-snapshots.sh`
- [x] coverage gates: `ferrocat-po` 97.53%, `ferrocat-icu` 96.85%
- [x] `node scripts/generate-collation-table.mjs --check` (including the Intl accent oracle)
- [x] `node scripts/check-published-numbers.mjs`
- [x] docs install and production build
- [x] PR benchmark regression profile: 14/14 scenarios

## Notes

- This changes the behavior of the existing default rather than adding a public enum variant. The public API snapshot remains unchanged, but the first write can reorder a code-point-sorted catalog.
- The checked-in 10k file-to-file benchmark report measures 101.005 MiB/s versus the previous 108.997 MiB/s, a 7.3% reduction. The result is non-noisy (1.31% CV, 4.95% span), and the README/home figures now point to it.
- An identical stripped fat-LTO consumer build (`codegen-units = 1`) grows from 666,768 to 749,360 bytes: +82,592 bytes (+12.39%), without an ICU collation dependency.
- Older strict FCL readers will reject the new `order=collated` header; Ferrocat continues to read their legacy tagless bytewise files.

Closes #259


swernerx:
Addressed the Greptile P1 from the review of `3ded325` in `8df9b1f`:

- secondary accent ranks are now derived from `Intl.Collator("en-US")` instead of Unicode code-point order;
- the compact key retains each accent's primary position, so identities such as `aá` and `áa` no longer collapse;
- precomposed and decomposed spellings retain canonical-equivalence behavior;
- the generator validates 90,951 accent pairs against the Intl oracle and Rust regressions cover message and context ordering, including the reported examples.

The subsequent `bef8c73` commit only corrects benchmark/ADR provenance and does not change runtime code.

PR #273, opened 2026-07-30
perf(po): allocation-focused catalog speedups and refreshed benchmark numbers

swernerx:
## Summary

Two profiling-driven optimization rounds over the catalog hot paths, each candidate implemented in an isolated worktree, validated for byte-identical output, and only kept when a sequential benchmark on the reference host (Apple M1 Ultra) showed a real improvement.

Published numbers move from this morning's baseline to:

| Workload | before | after |
|---|---:|---:|
| Full catalog update (`po-update/gettext-ui-de-10000`) | 112 MiB/s | **150 MiB/s** (+33%) |
| Update with new strings (`po-merge`) | 255 MiB/s | **291 MiB/s** (+14%) |
| Borrowed parse | 695 MiB/s | **725 MiB/s** |
| ICU-heavy update (internal) | ~97 MiB/s | ~144 MiB/s (+48%) |

The full-update lead over the fastest non-Ferrocat workflow update grows from 2.4x to 3.4x, and over GNU `msgmerge` to ~31x.

## Round 1 — allocation churn (samply showed 26% malloc + 15% memmove)

- `perf(po): drop throwaway sets in catalog merge helpers` — in-place dedup/compaction, `FxHashSet` for membership only, fast paths for empty/small inputs; new equivalence test against a naive reference across all size paths
- `perf(po): drop redundant buffers in catalog message import` — halves the catalog layer's own allocations on top of the parser; documents the in-place-collect size coupling between `CanonicalMessage` and the public `CatalogMessage`
- `perf(po): parse catalogs through the borrowed PO parser` — the catalog path consumes `parse_po_borrowed`, allocating each retained field exactly once; also fixes a pre-existing owned/borrowed parser divergence on irregular header blocks (validated with a 1.04M-input mutation differential, four regression tests)
- `perf(po): cut allocation churn in plural synthesis and PO export` — exact-size plural synthesis buffers, borrowed category tables (`&'static str`), presized serializer output
- `perf(po): cut per-message allocations in catalog export` — allocation-free placeholder-comment scan, origins rendered straight into the output buffer, per-operation gettext plural profile cache
- `perf(po): move each message once when applying collated order` — scatter through a scratch buffer instead of swap cycles, identity early-return

## Round 2 — after re-profiling the optimized build

- `perf(bench): ingest templates via the borrowed PO parser` — the compare harness parses `template.pot` zero-copy in the update and merge scenarios; conversion pinned byte-exactly against the previous owned conversion
- `perf(po): move merged message payloads instead of cloning` — identity matching as a separate pass, matched messages taken by value, `unchanged`/`changed` stats decided field-by-field instead of comparing against a retained clone (128-configuration differential: byte-identical output, stats, and diagnostics)
- `perf(po): continue collation prefixes on collision` — colliding 32-byte prefix buckets are re-keyed at offsets 32/64/96 before any full key is built (provably order-preserving; eliminates the full-key degeneration on Lingui-style offset-plural catalogs, 4000 -> 0 in the probe corpus). Same-msgid runs skip continuation entirely, which keeps ICU-heavy catalogs with repeated msgids at baseline speed.

## Verification

- `cargo test --workspace --all-features --locked`, `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `RUSTDOCFLAGS="-D warnings" cargo doc`, `cargo +1.93.0 check` (MSRV) — all green
- The compare suite's semantic-equivalence digests are unchanged against every recorded report, so output parity is verified against msgmerge, pofile, pofile-ts, polib, and Babel
- Published numbers are backed by three new checked-in reports at `c3caa27` and validated by `scripts/check-published-numbers.mjs` (18 values, 14 swept mentions); ADR 0026 keeps its historical collation measurements pinned to their original reports

## Notes for review

- The last four commits are unsigned (1Password signing agent was locked during the session); the earlier commits are signed as usual
- A pre-existing FCL bug surfaced during differential testing (plural entries round-tripped through FCL trip a spurious duplicate-id conflict on re-merge, reproducible on the baseline) — tracked separately, not addressed here


crates/ferrocat-po/src/api/plural.rs
```diff
@@ -607,14 +675,37 @@ pub(super) fn derive_plural_variable(
...
+    }
+}
+
+/// Renders already-ordered `(category, value)` branches, presizing the output so
+/// the buffer never has to grow.
+fn render_icu_plural(variable: &str, branches: &[(&str, &str)]) -> String {
+    // `{` + variable + `, plural,` + branches + `}`, where every branch costs
+    // ` ` + category + ` {` + value + `}`.
+    let capacity = variable.len()
+        + 10
```
swernerx: The fixed overhead is one byte short: `{` (1) + `, plural,` (9) + the final `}` (1) totals 11 bytes, not 10. With the current capacity, the final `out.push('}')` exceeds the requested capacity and can trigger exactly the buffer growth this optimization intends to eliminate for every synthesized ICU plural. Please change this to `+ 11`; a small regression assertion comparing the calculated capacity with the rendered length would also protect the no-growth contract.

swernerx (commented):
Change requested: the implementation and verification look strong overall, and all CI plus the local `ferrocat-po` test suite are green. I found one small but real issue in the allocation-focused hot path that should be fixed before merge.

crates/ferrocat-po/src/api/plural.rs
swernerx: Fixed in 70ad183 — the fixed overhead is indeed 11 bytes. The synthesis test now asserts `capacity() == len()` on the rendered output, which pins the no-growth contract; mutation-checked that the assertion fails against the previous `+ 10`.

PR #276, opened 2026-07-30
feat(po): preserve translator comments and opaque flags across catalog updates

### LINKED ISSUE #271: feat(po): preserve translator comments and opaque flags across catalog updates
## Problem

Ferrocat's high-level PO import currently collapses extracted comments (`#.`) and translator comments (`#`) into one canonical notes list, then exports that list as extracted comments. Per-entry PO flags are not retained by the canonical catalog model.

That is reasonable for catalog semantics, but it makes a source-first update lossy for hand-maintained PO metadata:

```po
# translator note
#. stale extractor note
#: src/Old.tsx
#, fuzzy, no-wrap
msgctxt "button"
msgid "Hello"
msgstr "Hallo"
```

A host normally wants to refresh extractor-owned metadata (`#.` and `#:`) while preserving translator-owned comments and opaque flags. Palamedes currently works around this by parsing the original PO, stripping translator comments and flags before the Ferrocat update, then parsing the rendered output and restoring them by `(msgctxt, msgid)`.

The workaround adds a full parse/serialize pass whenever any entry in a catalog carries this metadata.

## Desired direction

Provide lossless PO round-trip preservation for per-entry metadata that is outside Ferrocat's canonical catalog semantics:

- preserve translator comments separately from refreshed extracted comments;
- preserve flags opaquely without requiring Ferrocat to give `fuzzy`, `no-wrap`, or unknown flags semantic meaning;
- associate preserved metadata with the stable message identity through update and obsolete transitions.

This could be an internal PO round-trip layer around the canonical model rather than expanding the public semantic catalog shape.

## Acceptance criteria

- Translator comments remain translator comments (`#`) after a high-level PO update.
- Newly extracted comments replace or refresh extractor-owned comments without converting translator comments.
- Existing per-entry flags survive updates unchanged unless an explicit policy removes them.
- Unknown flags are preserved opaquely; no new semantic interpretation is required.
- Preserved metadata follows the same `(msgctxt, msgid)` identity and remains stable when an entry becomes obsolete.
- Repeating an unchanged update is byte-stable.
- New entries do not inherit metadata from unrelated identities.
- A downstream host no longer needs to strip and restore this metadata in a post-processing pass.

## Context

Palamedes carries the current workaround in `crates/palamedes/src/catalog_update.rs`; it was retained while removing the downstream sorting layer after the Ferrocat 3.0.1 collated-order release.

swernerx:
Closes #271

## Summary

High-level PO updates were lossy for hand-maintained metadata: translator-written `#` comments were flattened into extractor-owned `#.` comments, and per-entry `#,` flags were dropped entirely. Palamedes worked around this with a strip-and-restore post-pass around every update.

The canonical message model now carries `translator_comments` and `flags` as **internal-only, opaque round-trip state** attached to the `(msgctxt, msgid)` identity:

- **Import** keeps `#` comments separate from the `#.` pipeline and captures flags verbatim
- **Update** moves the previous entry's metadata onto the merged entry; extractor-owned `#.` and `#:` refresh as before; new entries never inherit metadata
- **Export** renders gettext canonical order: `#`, `#.`, `#@`, `#:`, then a single `#, flag1, flag2` line
- **Combine** (in-memory and file-based) adopts the metadata block of the definition whose translation owns the entry, mirroring the existing machine-metadata rule — never a cross-input union
- **Obsolete transitions** keep the metadata attached, and repeating an unchanged update is byte-stable (`!result.updated` pinned by test)
- **FCL** gains optional `tc=` (translator comment) and `f=` (flag) entry tags; canonical tag order is now `r, c, tc, f, o, lock, ai`. Files without the new tags are unaffected; files carrying them require a reader with this change (same additive-tag trade as `order=collated`)

## Semantics stay dropped — ADR 0027

This partially supersedes ADR 0023: the catalog layer still gives flags and the comment-kind split **no semantic meaning** — no `Fuzzy` status, no audit checks, no coverage counters. What changes is only that the metadata survives instead of being destroyed. ADR 0023 carries a supersession note; `docs/fcl-format.md` and the API overview document the new behavior. The public `CatalogMessage` is unchanged (API snapshot byte-identical).

## Notes for review

- One-time visible diff on first update after upgrade: previously flattened translator comments move back from `#.` to `#`, and previously dropped flags reappear. Downstream tooling that relied on Ferrocat silently clearing `fuzzy` will see it preserved now — that is the contract change this issue asks for.
- The catalog export writes `#, flag1, flag2` (GNU-style separator); the low-level `stringify_po` keeps its pre-existing `#, flag1,flag2` form — both parse identically, and the low-level round-trip is out of scope per ADR 0023/0027.
- A `.pot` header-block `#, fuzzy` (above `msgid ""`) is file-level, remains dropped, and verifiably does not leak onto the first message.

## Verification

Full workspace tests (+13 new covering every acceptance criterion in #271), fmt, clippy `-D warnings`, rustdoc `-D warnings`, MSRV check, docs build, public API snapshots **unchanged**, and the benchmark equivalence digest byte-identical to all committed baseline reports. Commits are unsigned (signing agent unavailable during the session).


crates/ferrocat-po/src/api/combine.rs
```diff
@@ -454,6 +459,16 @@ fn merge_combine_message(
         entry.message.machine = message.machine.clone();
     }
 
+    // Opaque translator metadata belongs to whichever definition ends up owning
+    // the entry value, and moves as one block. Unioning it across inputs would
+    // invent comments and flags no single catalog ever declared.
+    if translation_merge.changed
```
swernerx: `changed` is not equivalent to “the source definition owns the final translation”, and the fallback below is not gated by `matches_source`. For example, with `UseFirst`, an earlier non-empty translation with an empty metadata block retains its value, but a later conflicting (or empty) definition still supplies its comments/flags via the fallback. Plurals expose the broader problem: `changed` can be true after only filling/replacing some categories while the final map still does not match the source, yet the complete source metadata block is adopted for that composite value. This contradicts ADR 0027 and differs from the machine-metadata ownership handling immediately above. Please derive the transfer from actual ownership (`matches_source`), define the behavior for mixed plural results (the machine block is cleared there), and add regressions for a losing `UseFirst` definition and a partial plural merge.

swernerx (commented):
Change requested: the PO/FCL round-trip implementation and verification look solid overall, and all CI plus the local `ferrocat-po` suite are green. The combine ownership rule still has a correctness gap that can attach opaque metadata from a definition that did not win the entry.

crates/ferrocat-po/src/api/combine.rs
swernerx: Fixed in b358744 — the transfer now mirrors the machine-metadata ownership rule exactly: adopt only when `translation_merge.matches_source` (the merged value equals the incoming definition's translation), clear the block when `changed && !matches_source` (composite plural fill owned by no definition, same as the machine block), and the empty-target fallback is now gated on `matches_source` too, so a losing `UseFirst` definition can no longer supply metadata. Both requested regressions are in: `combine_catalogs_does_not_adopt_metadata_from_a_losing_definition` and `combine_catalogs_clears_metadata_for_composite_plural_values`.

PR #280, opened 2026-07-30
fix(coverage)!: classify fuzzy entries as incomplete

### LINKED ISSUE #278: fix(coverage): preserve fuzzy status in catalog coverage reports
## Problem

The original coverage issue #137 required fuzzy entries to be counted separately and not as translated. Ferrocat 3.1.0 currently does the opposite: the public catalog projection drops flags and the coverage test explicitly counts `#, fuzzy` as translated.

At the same time, Ferrocat 3.1.0 restored opaque translator comments and flags through catalog updates and FCL `tc=`/`f=` tags in #271. The information therefore survives high-level storage workflows, but is unavailable to `measure_catalog_coverage`.

Palamedes has to compensate by:

1. parsing the catalog through the format-neutral API for identities and translations;
2. parsing PO a second time through the low-level API;
3. building a fuzzy-key set;
4. overlaying that set on its own coverage loop.

Current workaround:

- https://github.com/sebastian-software/palamedes/blob/codex/ferrocat-3-1-cleanup/crates/palamedes-cli/src/commands/report.rs
- https://github.com/sebastian-software/palamedes/pull/484

This also prevents Palamedes from using Ferrocat's canonical plural completeness rule directly. Palamedes' projected `translated` boolean currently treats a plural as translated when any form is non-empty, while Ferrocat correctly treats a plural with any empty required form as incomplete.

## Proposed direction

Make coverage/review classification able to observe the semantic `fuzzy` state without requiring every public `parse_catalog` caller to pay for or expose all opaque PO metadata.

Possible shapes include:

- a file/content coverage API that parses with the required review metadata internally;
- an opt-in parse option that preserves review status;
- a dedicated coverage input projection;
- restoring `CatalogMessageStatus::Fuzzy` and `CatalogLocaleCoverage::fuzzy`.

The exact shape is open, but audit, coverage, and review should continue to share one status classifier.

## Acceptance criteria

- Active `#, fuzzy` PO entries do not count as translated.
- Active FCL entries carrying `f=fuzzy` behave identically.
- Per-locale coverage exposes a fuzzy count and optional fuzzy detail rows/source keys.
- Empty, partially empty plural, obsolete, missing, translated, fuzzy, and extra states remain mutually consistent across coverage/audit/review.
- The default high-throughput public parse path can continue skipping opaque metadata when review state is not requested.
- Palamedes can replace its manual join and second PO parse with `measure_catalog_coverage`.

## Relationship to existing work

This restores an acceptance criterion from #137 now that #271/3.1.0 preserve the underlying opaque flag through PO and FCL storage.

swernerx:
## Summary

- add `parse_catalog_for_review`, which keeps exact fuzzy identities without exposing arbitrary PO/FCL metadata
- restore the shared fuzzy status, coverage counter, audit diagnostic, and review behavior
- keep the default `parse_catalog` path metadata-free, including its FCL fast path
- move the CLI and benchmark coverage workflow to the shared classifier
- document the migration and the narrow ADR 0027 exception in ADR 0028

## Migration

`measure_catalog_coverage`, fuzzy-enabled `audit_catalogs`, and current review targets now require catalogs from `parse_catalog_for_review`. Programmatically built catalogs can opt in with `into_normalized_view_assuming_no_fuzzy`.

This is intentional: a catalog parsed through the default metadata-discarding path can't tell "no fuzzy flag" from "fuzzy state was dropped," so report APIs reject it instead of overstating completion.

## Testing

- `cargo fmt --all --check`
- published-crate packaging with `--locked --allow-dirty`
- docs install and production build
- public API snapshot check
- workspace llvm-cov run and coverage gate (`ferrocat-po` 97.55%, `ferrocat-icu` 96.85%)

Closes #278

BEGIN_COMMIT_OVERRIDE
fix(coverage): classify fuzzy entries as incomplete
END_COMMIT_OVERRIDE

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,13 @@ msgstr "world"
 
 For the common "merge fresh extracted messages into an existing catalog" workflow, `merge_catalog` is the lean Gettext-style entry point. For N-way catalog overlays and `msgcat`-style set operations, use `combine_catalogs`; when catalogs already live on disk, `combine_catalog_files` adds format inference and atomic output replacement around the same semantics. For release checks across a source catalog and target catalogs, use `audit_catalogs`. For dashboards or translator handoffs, use `measure_catalog_coverage` and `review_catalogs`. For application delivery, compile requested-locale artifacts with fallback and ICU diagnostics; use `compile_catalog_artifact_report` when host tooling also needs per-message resolution provenance without changing the runtime artifact shape.
 
+Coverage, audit, and review inputs should be parsed with
+`parse_catalog_for_review`. It returns a normalized catalog that retains only
+semantic fuzzy identities, so active `#, fuzzy` PO and `f=fuzzy` FCL entries do
+not count as translated. The ordinary `parse_catalog` path continues to skip
+opaque flags and translator comments for high-throughput consumers that do not
+request report state.
+
 Beyond the basics, Ferrocat exposes byte-oriented and allocation-light borrowed parsers for hot paths, ICU analysis and source/translation compatibility checks (`analyze_icu`, `compare_icu_messages`, `validate_icu_formatter_support`), ICU-aware pseudolocalization, semantic metadata normalization around `msgid + msgctxt`, and AI-translation metadata that high-level writers clear automatically once a human edits the text. ICU scope is MessageFormat v1; MessageFormat 2 is tracked as a future standard but is not a near-term target.
 
 ## What Ferrocat Gives You
--- a/api-snapshots/ferrocat-po.txt
+++ b/api-snapshots/ferrocat-po.txt
@@ -8,6 +8,7 @@ pub const ferrocat_po::diagnostic_codes::catalog::ALL: &[&str]
 pub const ferrocat_po::diagnostic_codes::catalog::DUPLICATE_METADATA: &str
 pub const ferrocat_po::diagnostic_codes::catalog::EMPTY_TRANSLATION: &str
 pub const ferrocat_po::diagnostic_codes::catalog::EXTRA_TRANSLATION: &str
+pub const ferrocat_po::diagnostic_codes::catalog::FUZZY_FLAG: &str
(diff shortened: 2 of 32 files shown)


swernerx (commented):
Read this against #278 and it lands where the issue asked. The part I like most is that `parse_catalog_for_review` restores the semantics without widening `CatalogMessage`: `OpaqueCapture::ReviewState` is a narrow, honest projection, and having the FCL path still validate the escapes in the tags it deliberately throws away is a nice touch. Making `measure_catalog_coverage` reject a metadata-discarding catalog instead of quietly reporting "no fuzzy" is the right call, and ADR 0028 explains why well enough that the migration reads as a decision rather than an accident.

Two things I'd like your read on, neither blocking.

`CatalogLocaleCoverage::fuzzy()` derives the count by subtracting the other four states from `total` instead of counting it. That is exact today, but `CatalogMessageStatus` is `#[non_exhaustive]`, so the next variant added there lands silently in the fuzzy bucket while `increment_status` ignores it just as silently. Since this PR already ships a breaking change, was there a reason not to add the counter field outright — or at least a `debug_assert` in `coverage_for_locale` that the buckets sum to `total`?

`translation_change_report` now skips everything that isn't `Translated`, so a translation that changed *and* got marked fuzzy no longer appears in the change details. That's arguably the most review-worthy case: somebody edited the string and flagged it for a second look. The coverage counters still surface it as fuzzy so nothing disappears outright, and the doc comment does say so — I just want to confirm it's the intended reading rather than a side effect of reusing the classifier.

CI is green across the board, the patch coverage holds, and the migration note in the description is clear about what callers have to change. This is an approve from me.

swernerx:
Followed up in e71bb4b. Fuzzy remains an incomplete coverage state, but translation value changes are tracked independently again. The regression now asserts the fuzzy rollup plus the change summary and detail row.

I kept the derived `fuzzy()` accessor for the semver-compatible report shape. A future `CatalogMessageStatus` variant can't land silently here because `increment_status` is an exhaustive match, so the compiler forces that decision.

PR #281, opened 2026-07-30
feat(icu)!: canonicalize runtime apostrophe quoting

### LINKED ISSUE #277: feat(icu): support policy-aware runtime apostrophe canonicalization
## Problem

`IcuSyntaxPolicy::RuntimeLiteralApostrophes`, added for #133, currently implements runtime-lenient parsing by replacing every apostrophe with `''` before strict ICU parsing.

That accepts natural-language apostrophes, but it is not equivalent to runtimes that keep ICU apostrophe quoting while tolerating unescaped literal apostrophes:

- `L'{title}` must keep `{title}` quoted; doubling the apostrophe makes it a live argument.
- `''` represents one escaped apostrophe; doubling both characters changes the parsed literal.
- `'#'` inside a plural branch must keep `#` quoted; doubling the apostrophes makes `#` active plural syntax.

Palamedes therefore still owns a context-sensitive canonicalizer and must apply it consistently to catalog `msgid`s, singular/plural translations, audit input, pseudolocalization input, and compiled-key derivation:

- https://github.com/sebastian-software/palamedes/blob/codex/ferrocat-3-1-cleanup/crates/palamedes/src/icu_text.rs
- https://github.com/sebastian-software/palamedes/blob/codex/ferrocat-3-1-cleanup/crates/palamedes/src/catalog_artifact/compile.rs
- https://github.com/sebastian-software/palamedes/pull/484

This leaves a shared ICU invariant outside Ferrocat and makes any new caller that skips the Palamedes pre-pass parse or hash a different message.

## Proposed direction

Extend the syntax-policy abstraction with context-sensitive runtime apostrophe handling rather than a global replacement. The same policy should own both parsing and canonical text derivation.

A public primitive such as `canonicalize_icu_with_policy` (exact naming open) would let downstream extraction/transform code derive the same canonical text and compiled ID that catalog audit and artifact compilation use.

Strict parsing should remain the default.

## Acceptance criteria

- Natural apostrophes such as `don't` parse as literal text under the runtime policy.
- Existing ICU quoting remains meaningful:
  - `L'{title}` keeps `{title}` literal.
  - `''` produces one apostrophe.
  - `'#'` inside plural branches keeps `#` literal.
- Canonicalization is idempotent.
- Audit, compatibility checks, artifact compilation, selected compilation, pseudolocalization, compiled-key derivation, and compiled-ID indexing can use the same policy.
- Inputs without apostrophes retain the current allocation-free/borrowed fast path where practical.
- Palamedes can remove `canonicalize_runtime_quoting`, `canonicalize_catalog_quoting`, and its compiled-key wrapper.

## Relationship to existing work

This is a correctness follow-up to #133, not a request for another parallel syntax-policy API.

swernerx:
## Summary

- Replace global apostrophe doubling with context-aware runtime ICU canonicalization.
- Add policy-aware canonical text, key derivation, and compiled-ID indexing while keeping strict defaults and `compiled_key` unchanged.
- Apply the same canonical semantics to audit, compatibility checks, compilation, selected compilation, and pseudolocalization.
- Document the runtime contract, stable-key behavior, and migration path in the public guides and ADRs.

## Breaking change

`RuntimeLiteralApostrophes` now canonicalizes both compiled IDs and artifact text. Callers that used this policy on Ferrocat 3.1.0 or earlier must regenerate runtime artifacts and `CompiledCatalogIdIndex` values together. Previously cached IDs don't match the canonicalized contract.

BREAKING CHANGE: RuntimeLiteralApostrophes now canonicalizes compiled IDs and artifact text. Regenerate runtime artifacts and CompiledCatalogIdIndex values created with Ferrocat 3.1.0 or earlier.

## Testing

- `cargo fmt --all --check`
- `cargo llvm-cov --workspace --all-features --locked --lcov --output-path target/lcov.info --ignore-filename-regex "crates/ferrocat-(bench|conformance)/"`
- `cargo llvm-cov report --json --summary-only --output-path target/coverage-summary.json --ignore-filename-regex "crates/ferrocat-(bench|conformance)/"`

Closes #277

BEGIN_COMMIT_OVERRIDE
feat(icu): canonicalize runtime apostrophe quoting
END_COMMIT_OVERRIDE


swernerx:
Fixed the report-policy divergence in bfaeaf8. `with_options` now copies the nested ICU options into the authoritative report field, while `with_icu_options` updates both values, so the last builder call wins without changing direct-field behavior. Regression coverage compares normal and report compilation for both full and policy-aware selected artifacts. The complete CI rerun is green.

swernerx (commented):
The canonicalizer is the good part here. Replacing the global `replace('\'', "''")` with a context-aware pass is exactly what #277 asked for, and `push_runtime_apostrophe` / `read_runtime_quoted_literal` get the cases that matter right: `''` stays one escaped apostrophe, a quote only opens before `{`, `}`, or a plural `#`, and re-emitting a quoted run re-doubles the inner apostrophes so the result is genuinely idempotent rather than idempotent-by-test. The `plural_depths` stack inheriting into nested arguments matches how ICU scopes `#`. And having selected compilation recompute each requested ID under the active policy and reject a mismatch, instead of quietly emitting a different ID, is the guard I'd want on this.

One thing I want resolved before this merges.

On main, `RuntimeLiteralApostrophes` only affected parsing. `compiled_key_for` always hashed the raw `msgid`, and `resolved.message` went into the artifact untouched. After this PR both are policy-canonical. So for anyone already compiling with `RuntimeLiteralApostrophes`, every message containing an apostrophe now gets a different compiled ID and different artifact text than 3.1.0 produced. Previously shipped artifacts and ID indexes stop matching, and the failure mode is a silent runtime lookup miss, not an error.

The commit is `feat(icu):` and the changelog entry reads as purely additive, so on a plain `^3` upgrade that break is invisible. Either mark it breaking (`feat(icu)!:` with a `BREAKING CHANGE:` footer) or add an explicit migration line saying artifacts and compiled-ID indexes built with the runtime policy on 3.1.0 or earlier must be regenerated. #280 already forces a major, so marking it is probably the cheaper option — but I'd rather the correctness of the release notes not depend on which PR lands first.

Two smaller things, take them or leave them.

`CompileCatalogArtifactReportOptions` still carries ICU options in two public places. bfaeaf8 keeps the builders in sync and the doc comment is honest about the remainder, but a caller assigning the fields directly can still end up with report compilation and artifact compilation on different policies. If a major is in flight anyway, dropping the top-level `icu_options` in favour of `options.icu_options` removes the trap instead of documenting it.

In `compile_catalog_artifact_from_source_keys`, `resolved.message` is canonicalized and then handed to `parse_icu_with_syntax_policy(..., IcuSyntaxPolicy::Strict)`. That's correct, but a plain `parse_icu` would say the same thing without making the next reader check whether the policy went missing.

So: changes requested on the semver and migration point only. Everything else above is optional.

swernerx:
Addressed the blocking release point in 5e988cd. The follow-up commit has a `!` plus a `BREAKING CHANGE:` footer, and the PR title/body now carry the same signal for squash merges. Both crate changelogs, the runtime guide, and ADR 0010 tell existing `RuntimeLiteralApostrophes` users to regenerate artifacts and `CompiledCatalogIdIndex` values together.

I left the duplicate public report field and the explicit strict parse call alone; those were optional cleanup suggestions and aren't needed for the migration fix.

PR #282, opened 2026-07-30
feat(catalog): add explicit PO/FCL conversion

### LINKED ISSUE #279: feat(catalog): add lossless cross-format PO/FCL conversion APIs
## Problem

Ferrocat has lossless PO and FCL codecs, but no public cross-format conversion API.

`combine_catalog_files` cannot fill this role:

- it resolves one `CatalogFileFormat`;
- that format selects both the input parser and output renderer;
- inferred input/output formats are required to match;
- an explicit `Fcl` format therefore makes a `.po` input get parsed as FCL rather than converting it.

Palamedes currently tries to implement `pmds catalog convert --to fcl` by calling the combine API with one input and `format: Fcl`:

- https://github.com/sebastian-software/palamedes/blob/codex/ferrocat-3-1-cleanup/crates/palamedes-cli/src/commands/catalog/convert.rs
- https://github.com/sebastian-software/palamedes/blob/codex/ferrocat-3-1-cleanup/crates/palamedes/src/catalog_combine.rs

That is both the wrong ownership boundary and a functional mismatch with the 3.1.0 API.

Ferrocat 3.1.0 already has the internal pieces: canonical PO/FCL import/export, deterministic collation, atomic file replacement, and preservation of translator comments/opaque flags through FCL `tc=`/`f=` tags.

## Proposed direction

Add explicit content and file conversion APIs, for example:

- `convert_catalog`
- `convert_catalog_file`
- options with distinct source and target formats/modes

The file API should validate and render fully before atomically replacing the output path, following the existing update/combine file APIs.

Palamedes would retain config discovery, locale iteration, CLI presentation, and deciding which catalogs to convert.

## Acceptance criteria

- PO → FCL and FCL → PO are explicitly supported without pretending both sides use one format.
- Round trips preserve message/context identity, singular and ICU plural translations, extracted and translator comments, origins/scopes, opaque flags including `fuzzy`, obsolete metadata, and valid machine metadata.
- Output uses the target format's canonical deterministic ordering and rendering.
- Locale/source-locale and semantic-mode validation happen before writing.
- Unsupported or lossy conversions return an error without changing an existing output file.
- File conversion uses atomic replacement and supports different input/output paths safely.
- Focused tests cover PO → FCL → PO, including `tc=`/`f=` metadata and plurals.
- Palamedes can implement `pmds catalog convert` as thin host/config orchestration.

swernerx:
## Summary

- add explicit in-memory and file conversion APIs for PO and FCL, with independent source and target modes
- preserve shared message metadata and reject locale, semantic, and FCL representability mismatches before atomic writes
- cover round trips, same-path conversion, placeholder metadata, and failure safety, and document the durable boundary in ADR 0029

## Validation

- `cargo fmt --all --check`
- `cargo llvm-cov --workspace --all-features --locked --lcov --output-path target/lcov.info --ignore-filename-regex 'crates/ferrocat-(bench|conformance)/'`
- `cargo llvm-cov --workspace --all-features --locked --json --summary-only --output-path target/coverage-summary.json --ignore-filename-regex 'crates/ferrocat-(bench|conformance)/'`

Closes #279


swernerx (commented):
This one reads well. Splitting source and target modes instead of overloading a single `CatalogFileFormat` is the right fix for the ownership problem described in #279, and finishing all parsing, validation, and rendering before `atomic_write` gets you same-path conversion and failure safety as a property of the control flow rather than as a convention someone has to remember. `validate_catalog_fcl` catching identities that PO can legally hold apart but FCL cannot is a good catch: an active plus obsolete pair would otherwise have collapsed silently on export.

`po_fcl_po_roundtrip_preserves_shared_message_metadata` is doing real work too — `tc=`/`f=`, custom flags, ICU plurals, obsolete entries, lock and ai metadata, and it asserts the header drop rather than leaving it implicit.

Three follow-up thoughts, none of them blocking.

Dropping `headers`, `file_comments`, and `file_extracted_comments` at the FCL boundary is documented in ADR 0029 and covered by the test, but it happens silently. `CatalogConvertResult` already carries `diagnostics`; an info diagnostic naming what was dropped would let a CLI tell someone their `Project-Id-Version` and `Last-Translator` are gone, instead of them finding out later.

`validate_catalog_fcl` only runs from `finish_catalog_conversion`. The same collision can still reach FCL output through `update_catalog` and `combine_catalogs`. That's pre-existing and outside this PR's scope, but the check now exists, so wiring it into the FCL renderer would cover all three paths. Happy for that to be its own issue.

Conversion parses with `strict = false` and there's no way to opt in. Reasonable as a default for a format-shifting operation, but a `with_strict` on `ConvertCatalogOptions` would let a CLI refuse to convert a catalog with broken ICU rather than faithfully carrying the breakage across.

This is an approve from me.

swernerx:
Opened #283 for the shared FCL representability gap. The fix should move serialized `(id, ctxt)` validation to the common FCL export boundary so conversion, update, and combine can't drift.

I'm keeping metadata-drop diagnostics and an optional strict conversion mode out of this PR. The metadata boundary is already explicit, while strict ICU conversion needs a separate product contract rather than just another boolean.

